PR #1, opened 2024-08-10
First version of @alloy-js/java

IlluzionzDev:
Implements a first working version of the Java implementation of Alloy.

Main Features Implemented:
- Dependency and import management to source files
- Creating a basic maven project
- Package management and declarations
- Declaring external libraries to be imported into files
- Basic java components (see below)

Main java features implemented:
- Class declarations
- Enum declarations
- Interface declarations
- Annotations (Not declaring a new custom annotation), just annotating objects
- Variable declarations
- Method declarations


packages/core/src/context.ts
```diff
@@ -21,15 +21,15 @@ export interface ContextProviderProps<T = unknown> {
 export function useContext<T>(context: ComponentContext<T>): T | undefined {
   // context must come from a parent
   let current = getContext();
-  while (current) {
```
bterlson, line 24: These changes shouldn't be necessary (or if they are, are masking some other bug I suspect). Can you say more about why this is needed?

packages/java/.prettierignore
```diff
@@ -0,0 +1,2 @@
+# Can't format JSX sadly :( Need a fork.
```
bterlson, line 1: shouldn't this be covered by the prettier ignore in the root dir?

packages/java/src/components/Annotation.tsx
```diff
@@ -0,0 +1,20 @@
+import { code, mapJoin, Refkey, refkey } from "@alloy-js/core";
+import { Reference } from "./Reference.js";
+import { Child, Children } from "@alloy-js/core/jsx-runtime";
+import { collectNamedArguments } from "../arguments.js";
+
+export interface AnnotationProps {
+  type: Child;
+  value?: Child; // If one argument, provide this
+  arguments?: Record<string, Child>; // Arg name to value, provide if many named arguments
```
bterlson, line 9: arguments doesn't feel descriptive to me. Maybe `values` would be better? Ideally this could be `value?: Child | Record<string, Child>` but some objects can be children (refkeys, refs), so that check is a little risky I guess.

packages/java/src/components/Class.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { collectArguments } from "../arguments.js";
+
+export interface ClassProps extends DeclarationProps, ObjectModifiers {
+  accessModifier?: AccessModifier;
+  extends?: Children;
+  implements?: Children;
+}
+
+export function Class(props: ClassProps) {
+  if (Array.isArray(props.extends) && props.extends.length > 1) {
```
bterlson, line 15: This check isn't good because an array of children doesn't necessarily mean that it represents multiple extends, it could just be an array of strings (e.g. something like `["foo", "<", "T", ">" ]`)

packages/java/src/components/Class.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+  const name = useJavaNamePolicy().getName(props.name, "class");
+  const extendExpression = props.extends ? code` extends ${props.extends}` : "";
+  const collectedInterfaces = collectArguments(props.implements)
+  const implementsExpression = props.implements ? code` implements ${collectedInterfaces}` : "";
+  const modifiers = collectModifiers(props);
+  return (
+    <Declaration {...props} name={name}>
+      {collectAccessModifier(props.accessModifier)}{modifiers}class <Name />{extendExpression}{implementsExpression} {"{"}
+        <Scope name={name} kind='class'>
```
bterlson, line 27: This is something the typescript components need too ☺️

packages/java/src/components/Declaration.tsx
```diff
@@ -0,0 +1,37 @@
...
+ * access modifier so we can manage access
+ */
+export function Declaration(props: DeclarationProps) {
+  const sym = createJavaSymbol(props);
+  return <CoreDeclaration symbol={sym}>
+    {props.children}
+  </CoreDeclaration>;
+}
+
+export function createJavaSymbol(props: DeclarationProps) {
```
bterlson, line 23: Consider moving this to symbols.ts

packages/java/src/components/PackageDirectory.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+  const packagePath = sourceDirectory?.path + "/" + packageName;
+  const packageContext: PackageDirectoryContext = {
+    scope,
+    path: packagePath,
+    name: packageName,
+    qualifiedName: fullyQualifiedPackageName
+  };
+
+  function ChildComponent() {
```
bterlson, line 53: This needs a more descriptive name

packages/java/src/object-modifiers.ts
```diff
@@ -0,0 +1,37 @@
...
+  "package-private": ""
+}
+
+export function collectAccessModifier(accessModifier?: AccessModifier) {
+  return accessModifier ? accessModifierLookup[accessModifier] : "";
+}
+
+export function collectModifiers<T extends ObjectModifiers>(props: T) {
+  const modifiers = [];
+  if (props.abstract) modifiers.push("abstract");
```
bterlson, line 27: Java... 🤷‍♀️

bterlson (changes requested):
This is AMAZING!!! Nice work! The sample looks really good.

packages/core/src/context.ts
IlluzionzDev: Ahh yea, I was using Alloy in typespec incorrectly, and the errors were pointing here. Just turns out it wasn't actually initing the context properly, forgot to revert this back.

packages/java/src/components/Annotation.tsx
IlluzionzDev: I think I'll implement it as just `value?: Record<string, Child>` and if there is only one key just pass the value of that key. In Java any argument going to an annotation is named anyway, just if there's only one argument you don't have to specify the name.

IlluzionzDev:
I've added a few commits to do some refactoring for some recent changes on typescript. I've also brough in the prettier changes, which is why there are random ' to " changes etc (I had to run `pnpm format`). I had to fix the type export for typescript aswell as the sample project wasn't building.

So just a bit of cleanup changes

PR #8, opened 2024-08-14
Migrate to pnpm catalogs

timotheeguerin:
Allows us to use the same version across packages without extra tools to make sure they are. Reducing diffs when upgrading

### DIFF
--- a/packages/babel-plugin-jsx-dom-expressions/package.json
+++ b/packages/babel-plugin-jsx-dom-expressions/package.json
@@ -17,19 +17,19 @@
     "test:coverage": "pnpm run build && jest --coverage --no-cache"
   },
   "dependencies": {
-    "@babel/helper-module-imports": "7.18.6",
-    "@babel/plugin-syntax-jsx": "^7.18.6",
-    "@babel/types": "^7.20.7",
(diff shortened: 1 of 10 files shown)


packages/babel-plugin-jsx-dom-expressions/package.json
```diff
@@ -17,19 +17,19 @@
...
-    "validate-html-nesting": "^1.2.1"
+    "@babel/helper-module-imports": "catalog:",
+    "@babel/plugin-syntax-jsx": "catalog:",
+    "@babel/types": "catalog:",
+    "html-entities": "catalog:",
+    "validate-html-nesting": "catalog:"
   },
   "peerDependencies": {
-    "@babel/core": "^7.20.12"
+    "@babel/core": "^7.24.7"
```
timotheeguerin, line 27: should that not be catalog?

timotheeguerin: hhm I guess it make sense you have a wider range for peerDependencies

timotheeguerin: i'll keep those separate for now

PR #10, opened 2024-08-21
@alloy-js/java: Implement Generics, better maven project handling, other misc changes

IlluzionzDev:
Implements a few features for @alloy-js/java:

- Able to declare generic types on components that support them
- Better able to handle config options for maven pom.xml files (Gradle support will come later down the line)
- Changes access modifiers to be boolean props instead of enums
- Enum declaration auto declare delimiter (',' or ';' depending on position in declaration)

Also includes a fix in the CI/CD when deploying via pkg-pr-new to run `pnpm pack` and properly support pnpm catalogs.


packages/java/src/components/Enum.tsx
```diff
@@ -37,9 +42,15 @@ export interface EnumMemberProps extends ObjectModifiers {
...
+  const enumScope = useScope() as JavaEnumScope;
+
   const collectedArgs = collectArguments(props.arguments);
   const args = props.arguments ? code`(${collectedArgs})` : "";
   const name = useJavaNamePolicy().getName(props.name, "enum-member");
 
-  return code`${name}${args}`;
+  // Handle position in enum scope to determine delimiter
+  enumScope.members.push(name);
+  const delimiter = memo(() => enumScope.getDelimiter(name));
```
bterlson, line 53: Use `mapJoin` in the enum when members are provided as opposed to handling delimiter in the enum member element. Try to keep the scope dedicated to symbol tracking information.

bterlson (approved):

packages/java/src/components/Enum.tsx
IlluzionzDev: No worries, I've reverted the change

PR #11, opened 2024-08-23
update typescript component exports to reference js instead of jsx

chrisradek:
Noticed some issues when trying to actually import the package due to referencing `jsx` files.


chrisradek:
Closed by https://github.com/alloy-framework/alloy/commit/7d274fd4a1f437a62dd33999738a9d66ac678bd9

PR #14, opened 2024-08-28
Support async function declarations

joheredi:
Adds an `async` prop to function declaration and handles the return type to be wrapped by a Promise

### DIFF
--- a/packages/typescript/test/function-declaration.test.tsx
+++ b/packages/typescript/test/function-declaration.test.tsx
@@ -38,6 +38,24 @@ it("can be a default export", () => {
     `);
 });
 
+it("can be an async function", () => {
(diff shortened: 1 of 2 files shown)


packages/typescript/test/function-declaration.test.tsx
```diff
@@ -38,6 +38,26 @@ it("can be a default export", () => {
     `);
 });
 
+it("can be an async function", () => {
+  expect(toSourceText(<FunctionDeclaration async export name="foo" />)).toBe(d`
+    export async function foo(): Promise<void> {
```
bterlson, line 43: Only adding the explicit return type when the function is async is odd to me? Can we just leave it off and let the compiler infer like the non-async case?

joheredi: Yeah this is a good point

PR #19, opened 2024-09-06
Initial commit of alloy-js/csharp

jhendrixMSFT:
Includes basic functionality including namespaces, classes and members, enums and members, and using directives with refkey support.


packages/csharp/src/components/Class.tsx
```diff
@@ -0,0 +1,110 @@
...
+  // this creates a new scope for the class definition.
+  // members will automatically "inherit" this scope so
+  // that refkeys to them will produce the fully-qualified
+  // name e.g. Foo.Bar.
+  const thisClassScope = symbols.createCSharpMemberScope(scope.binder, scope, thisClassSymbol, "class");
+
+  return <core.Declaration symbol={thisClassSymbol}>
+      {csharp.getAccessModifier(props.accessModifier)}class <base.Name />{props.children ? (
+        <>
+          {"\n{"}
```
bterlson, line 31: I think you should be able to include just an empty line? If not, that's a bug.

bterlson: You can also consider a <Block> component which adds the braces for you, slightly more characters than {"{"} but less annoying to type maybe.

packages/csharp/src/components/Class.tsx
```diff
@@ -0,0 +1,110 @@
...
+  });
+
+  // this creates a new scope for the class definition.
+  // members will automatically "inherit" this scope so
+  // that refkeys to them will produce the fully-qualified
+  // name e.g. Foo.Bar.
+  const thisClassScope = symbols.createCSharpMemberScope(scope.binder, scope, thisClassSymbol, "class");
+
+  return <core.Declaration symbol={thisClassSymbol}>
+      {csharp.getAccessModifier(props.accessModifier)}class <base.Name />{props.children ? (
```
bterlson, line 29: I might add `{!props.children && ";"}` at the start to avoid the massive ternary.

packages/csharp/src/name-policy.ts
```diff
@@ -0,0 +1,37 @@
...
+
+// the context in which the name policy should be applied
+export type CSharpElements =
+  | "class"
+  | "constant"
+  | "enum"
+  | "enum-member"
+  | "function"
+  | "interface"
+  | "member"
```
bterlson, line 12: class-member might be more descriptive? Unsure of common C# naming conventions so def follow those.

bterlson (approved):

packages/csharp/src/components/Class.tsx
jhendrixMSFT: Empty line does work.

What's the idiom here? Having a blank line looks weird to me but if that's what folks do then I'll follow suit.

PR #40, opened 2024-10-29
Add slots and replace/rename functionality.

bterlson:


packages/java/src/javaslots.ts
```diff
@@ -0,0 +1,35 @@
...
+    artifactId +
+    "." +
+    transformJavaFqn(pkg) +
+    "." +
+    javaFileName.replace(".", "_") +
+    (memberName ?
+      isStatic ? "_" + memberName
+      : "." + memberName
+    : "");
+  console.log("Resolved Java FQN", fqn);
```
jongio, line 20: This `console.log` looks like debug output that shouldn't ship. Remove it before merge.

packages/java/src/javaslots.ts
```diff
@@ -0,0 +1,35 @@
...
+  return resolveFQN(fqn);
+}
+
+function transformJavaFqn(input?: string): string {
+  if (!input) return "";
+  const parts = input.split(".");
+  return parts
+    .map((part, index) => {
+      if (index < parts.length) {
+        return parts.slice(0, index + 1).join("_");
```
jongio, line 30: The condition `if (index < parts.length)` is always true when iterating with `.map()` (index goes from 0 to length-1). The `else` branch is unreachable dead code. You can remove the conditional and just return the `parts.slice(0, index + 1).join("_")` directly.

Also worth adding a doc comment explaining what this function produces. For input `"com.example.service"` it outputs `"com.com_example.com_example_service"` (cumulative underscore-joined prefixes). That behavior isn't obvious from reading the code.

packages/core/src/binder.ts
```diff
@@ -854,18 +858,24 @@ export function createOutputBinder(options: BinderOptions = {}): Binder {
...
         waitingScopeNames.set(scope, new Map());
       }
       const waiting = waitingScopeNames.get(scope)!;
-      waiting.set(name, scopeRef);
+      const key = name.replace(/\./g, "_");
+      if (waiting.has(key)) {
+        return waiting.get(key) as Ref<TScope | undefined>;
+      }
+
+      const scopeRef = shallowRef<OutputScope | undefined>(undefined);
```
jongio, line 877: Line 873 defines `const key = name.replace(/\./g, "_")` but this line uses `name.replace(/\./g, "_")` again instead of reusing `key`. Minor inconsistency.

packages/core/test/components/slot.test.tsx
```diff
@@ -122,12 +116,55 @@ it("can rename", () => {
...
+          <MyFunctionComponent name="bar" />
+        </Scope>
+      </SourceFile>
+    </Output>,
+  );
+
+  console.log(tree.contents[0].contents);
+});
+
+it("can rename the same thing, last wins", () => {
```
jongio, line 161: This test (and the "can rename the same thing, last wins" test below it) ends with `console.log(...)` but has no `expect()` assertions. The test will always pass regardless of actual output. Consider adding assertions that verify the rename actually took effect in the rendered tree.

jongio (commented):
A few issues to address:

**javaslots.ts** has a debug `console.log` that should be removed before merge, and `transformJavaFqn` has dead code (the `if (index < parts.length)` condition is always true in a `.map()` callback).

**binder.ts** has a minor inconsistency where a computed `key` variable is defined but then `name.replace(...)` is used directly on the next line instead.

**Tests**: The rename tests log output to the console but don't assert on it. These tests will pass regardless of actual behavior.

Also noting: this PR currently has merge conflicts and is missing a changeset (per the chronus bot).

PR #46, opened 2024-12-22
feat: improve `ts.ValueExpression` conversion

TomerAberbach:
This handles two new cases:
1. Strings that have characters that need to be escaped
2. BigInts


TomerAberbach:
I think some of the CI stuff is not related to my change, but I went ahead and at least fixed the formatting thing

PR #52, opened 2025-02-11
Add JSDoc components to TS

joheredi:
* Add JSDoc component
* Wraps declaration in JSDoc to add if provided
* Support for function and class method parameters JSDoc


packages/typescript/src/components/EnumDeclaration.tsx
```diff
@@ -11,11 +12,31 @@ import { useTSNamePolicy } from "../name-policy.js";
...
 export interface EnumDeclarationProps extends BaseDeclarationProps {
   /**
-   * A JS object representing the enum member names and values.
+   * The members of the enum.
+   */
+  members?: Record<string, string | number | EnumMemberDescriptor>;
+  /**
+   * Documentation for the enum.
    */
-  jsValue?: Record<string, string | number>;
```
joheredi, line 18: @bterlson This would be a breaking change to allow values or descriptor. What do you think?

bterlson:
@joheredi this PR needs to be updated now with the new JSDoc components.

PR #55, opened 2025-02-17
Reactivity and control flow improvements

bterlson:
This PR fixes many issues related to reactivity, such as cleanup functions not being called, effects not being observed, and so forth. It also introduces an element cache during the rendering phase, which will reuse elements if they have not changed on a reactive re-render.

This foundation enables mapJoin to become lazy, where adding a new element to the array being mapped will not cause previous items to re-render. Currently, this is only true when appending items - any items after a changed element will be re-rendered. In the future this can be improved. In order to support this, **`mapJoin` takes a significant breaking change** - the array or map to be mapped must be wrapped in a callback - `mapJoin(someArray, ...)` becomes `mapJoin(() => someArray, ...)`. 

Additionally, this PR adds a `For` component which allows for using mapJoin ergonomically within the context of a JSX template. It also adds a `Switch` component for conditionally rendering elements. Combined, these enable moving significant amounts of logic into the template in a way that is still readable, and without requiring any complex gymnastics to leverage reactivity.


packages/core/src/render.ts
```diff
@@ -302,47 +305,69 @@ function appendChild(
...
   const lines = str.split("\n");
   return [lines[0], ...lines.slice(1).map((line) => indent + line)].join("\n");
 }
-type NormalizedChild = string | (() => Child | Children) | NormalizedChild[];
+type NormalizedChild =
+  | string
+  | (() => Child | Children)
+  | CustomContext
+  | NormalizedChild[]
+  | boolean
```
bterlson, line 368: This change not needed (brought over from formatting branch)

packages/core/src/utils.ts
```diff
@@ -55,18 +62,116 @@ export function mapJoin<T, U, V>(
...
+
+  function mapMap(item: [T, U]) {
+    return cb(item[0], item[1]);
+  }
+
+  function isIterable<T>(x: unknown): x is Iterable<T> {
+    return typeof (x as any).next === "function";
+  }
+  /*
   const mapped: (V | string)[] = [];
```
bterlson, line 175: delete

packages/core/src/jsx-runtime.ts
```diff
@@ -83,46 +103,98 @@ export function memo<T>(fn: () => T, equal?: boolean): () => T {
     // eslint-disable-next-line @typescript-eslint/no-unused-expressions
     (!equal || prev !== res) && (o.value = res);
     return res;
-  });
-  return () => o.value;
+  }, undefined as T);
+  const val = () => o.value;
+  (val as any).fn = fn;
```
bterlson, line 108: remove this

joheredi (approved):
:shipit:

PR #56, opened 2025-02-28
Add formatting and a bunch of new components

bterlson:
This PR implements formatting by allowing templates to include formatting commands from Prettier. For example, a template can include `<group>` which is a set of strings to try to keep on one line if possible, and various kinds of breaks that say how the group should break when it doesn't fit.

To support this, a number of additions and changes (many of them breaking!) have been implemented:

## JSX Transform

* The JSX transform now strips all leading and trailing whitespace in text nodes. Additionally, all spaces and line breaks in the middle of a text node are replaced with a space. Auto-indenting behavior and the notion of "significant" linebreaks are removed. Indents are now explicit either via the `indent` intrinsic or `Indent` component. All significant linebreaks must now be added via `<hbr />`, `<sbr />`, `<lbr />`, or `<br />`.
* As a result of this change, the prettier plugin is no longer needed, the standard JSX formatter works fine (though could be optimized to avoid some annoying patterns).

## Alloy core

* Various listing elements get a set of props for controlling the list behavior - `joiner`, `ender`, and boolean props for common joiner punctuation (`comma`, `semicolon`) and line breaks (`hard`, `soft`).
* The `For` component skips falsy values.
* The `Indent` component now doesn't include an ending line break by default, though it may be requested with `trailingBreak`. Use `trailingBreak` when you are indenting a block with an ending sigil (e.g. `}`), otherwise blocks should not end with a line break to allow subsequent blocks to set the indent level for themselves).
* Added `List` component which allows joining its child elements with a joiner/ender.
* The `Output` component now takes `printWidth`, `useTabs`, and `tabWidth` options for controlling the default formatting of all outputs. **The `indent` prop has been removed.**
* Added `Show` component for conditionally rendering some block of code (similar to the `someCondition && ...` pattern, but can look nicer in some cases.
* The `SourceFile` component now takes  `printWidth`, `useTabs`, and `tabWidth` options for controlling the default formatting for that specific source file.
* Added `Wrap` component for conditionally wrapping a component with other components.
* **Removed `indent` context**. It is no longer needed with indenting handled entirely by the printer.
* The JSX runtime now supports creating intrinsic elements. Formatting commands are created via intrinsic elements.
* Implemented `splitProps` to split a props object into two or more objects while handling reactivity. Useful for plumbing subsets of props into different child components.
* The renderer added the notion of a new type of Render Text Tree node which has a `print` callback that is called by the printer to get the rendered text for that node. Currently used for implementing formatting for intrinsic elements.
* Added `sti` function as peer for `stc` to create intrinsic elements. Probably not needed to be called by users as we should export all the intrinsic elements from `@alloy-js/core/stc`.
* Added `preserveFragments` option to `children` and `childrenArray`, which allows for using fragments in children as a list item container.

## JSON

* Now emits pretty code.

## TypeScript 

* Now emits pretty code
* Added `CommaList` component for creating lists separated by commas and soft or hard breaks (useful for object literals, arrays and such).
* Added `JSDoc`, `JSDocComment`, `JSDocParagraph`, and `JSDocExample` components for emitting nicely formatted JSDoc comments.
* Added `StatementList` component for creating lists separated by semicolons and hard breaks.

## Todo

* Documenting new components
* Formatting Java and C#


bterlson:
@joheredi gave this a once over, found that it seems to work and the breaking changes, while extensive, are fairly easy to handle.

packages/java/src/components/Enum.tsx
```diff
@@ -1,40 +1,56 @@
...
+export interface EnumMemberListProps {
+  children: Children;
+}
+
+/**
+ * Create a list of enum members, joining with a comma and a hardline, and
+ * ending with a semicolon.
+ */
+export function EnumMemberList(props: EnumMemberListProps) {
+  return <List children={props.children} comma hardline ender=";" />;
```
srnagar, line 55: Nice! Pretty concise to define delimiter, new line and ending. What's the difference between all the `*line` props in `BaseListProps` - hardline, literalline, line and softline? When should they be used?

packages/java/src/components/ImplementsClause.tsx
```diff
@@ -0,0 +1,19 @@
...
+export interface ImplementsClauseProps {
+  interfaces: Children[] | undefined;
+}
+
+/**
+ * The implements clause for a class, interface, or enum.
+ */
+export function ImplementsClause(props: ImplementsClauseProps) {
+  return (
+    <Show when={props.interfaces !== undefined && props.interfaces.length > 0}>
```
srnagar, line 12: `Show` is pretty neat!

packages/java/src/components/MavenProject.tsx
```diff
@@ -23,129 +30,257 @@ export function PomFile({ projectConfig }: PomFileProps) {
...
-  return <SourceFile path="pom.xml" filetype="xml">
+  return (
+    <SourceFile path="pom.xml" filetype="xml">
+      <PomFileRoot>
+        <hbr />
+        <List
+          joiner={
+            <>
+              <hbr />
+              <hbr />
```
srnagar, line 41: What's `hbr` and why do we need it twice here?

srnagar (commented):
The Java package changes look great to me! Thanks for refactoring and updating the code to support the new formatting.

packages/java/src/components/Enum.tsx
bterlson: Need to add docs for this! But basically:

* Hardline - always a line break, use for statements and such.
* Softline - only a line break when needed, use for things like breaks after opening parens or before closing parens.
* line - a line break when needed, otherwise a space. Use for things like argument lists which are separated by a comma, space.
* literalline - a linebreak which doesn't indent on the next line (rarely used, for things like template literals).

packages/java/src/components/MavenProject.tsx
bterlson: Hardbreak (essentially, "\n\n"). I don't know except that the test expects it.

PR #69, opened 2025-03-27
Support Source File header

joheredi:


packages/typescript/src/components/Prose.tsx
```diff
@@ -1,15 +1,15 @@
...
 
-export interface JSDocParagraphProps {
+export interface Prose {
   children: Children;
 }
 
 /**
- * Create a paragraph in a JSDoc comment. The text of this component is word
- * wrapped.
+ * Creates a prose element. This is a special element that breaks the children into
```
bterlson, line 9: Create a block of text which will break once a word exceeds the configured line width. The children are expected to be strings, and a `<br />` is added between each word.

packages/typescript/src/components/JSComment.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+export interface JSCommentProps {
+  children: Children;
+}
+
+/**
+ * A JS Comment. The children are rendered as a prose element, which means that they
+ * are broken into multiple lines
+ */
+export function JSComment(props: JSCommentProps) {
```
bterlson, line 12: I think this should be called after the grammar production, which is single line comment. `SingleLineCommentBlock` perhaps.

packages/typescript/src/components/JSComment.tsx
```diff
@@ -0,0 +1,22 @@
...
+  return (
+    <>
+      //{" "}
+      <align string="// ">
+        <Prose>{props.children}</Prose>
+      </align>
+      <hbr />
```
bterlson, line 19: It should not end in a line break (as it is if I put this in a <List> I'll get an extra line break).

bterlson (changes requested):
Couple comments, overall very nice.

bterlson:
Oh also make sure to add the Prose element to the changelog.

.chronus/changes/feature-source-file-header-2025-2-27-1-49-36.md
```diff
@@ -0,0 +1,10 @@
+---
+changeKind: feature
+packages:
+  - "@alloy-js/core"
+  - "@alloy-js/typescript"
+---
+
+- Support SourceFile header. This PR adds the following components:
```
bterlson, line 8: In future lets add separate changelog files for each change, this will mangle the changelog entries (Prose is not added to core for example).

bterlson (approved):

PR #70, opened 2025-03-27
Support optional chaining

joheredi:
- Support optional chaining in MemberChainExpression
-  Expose getSymbolForRefkey on the binder to lookup a symbol by name,

### DIFF
--- a/packages/core/src/binder.ts
+++ b/packages/core/src/binder.ts
@@ -337,6 +337,12 @@ export interface Binder {
   ): Ref<TScope | undefined>;
 
   /**
+   * Gets a symbol by its refkey.
+   * @param ref - The refkey to resolve.
+   */
+  getSymbolForRefkey(ref: Refkey): OutputSymbol | undefined;
+
+  /**
    * Resolve a fully qualified name to a symbol. Access a nested scope by name
    * with `::`, a nested static member with `.` and a nested instance member
    * with `#`.
@@ -432,6 +438,7 @@ export function createOutputBinder(options: BinderOptions = {}): Binder {
     instantiateSymbolInto,
     findSymbolName,
     findScopeName,
+    getSymbolForRefkey,
     resolveFQN: resolveFQN as any,
     globalScope: undefined as any,
   };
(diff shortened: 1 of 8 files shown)


packages/typescript/test/member-chain.test.tsx
```diff
@@ -119,6 +130,153 @@ it("works without function call expressions", () => {
       { printWidth: 5 },
     ),
   ).toBe(d`
-    z.string.min
+    z.string
+    .min
```
bterlson, line 134: If a linebreak is introduced here (which is fine), the next line should be indented.

packages/typescript/src/components/MemberIdentifier.tsx
```diff
@@ -0,0 +1,66 @@
...
+  isRefkey,
+  OutputSymbol,
+  Refkey,
+  useBinder,
+} from "@alloy-js/core";
+
+/**
+ * Common props shared by both variants
+ */
+export interface CommonMemberExpressionProps {
```
bterlson, line 12: I don't like how MemberIdentifier is concerned with both rendering the name of the symbol it is provided and exposing a `nullish` prop so that `MemberChainExpression` can pick it up. This feels like a conflation of concerns, and is a bit confusing (how can a member identifier be nullish??).

Maybe better would be to have a component like `MemberExpressionPart` which has `nullish` on it. As a baseline experience, when you need a ?., you have to wrap in that component (where today you just pass nullish).

But I still have a vague worry that something is missing. This does not work for optional calls for example. Hmm. Need to ponder.

packages/typescript/src/components/Interface.tsx
```diff
@@ -41,29 +49,38 @@ export interface InterfaceMemberProps {
...
       <>
         [{props.indexer}]: {type}
       </>
     );
   } else {
-    return (
+    const memberContent = (
       <>
         {readonly}
-        {namer.getName(props.name!, "interface-member")}
```
bterlson, line 64: Now that symbols are being created, the name needs to use the `MemberName` component (because it may be deconflicted with other symbols). But don't worry about this, I'm adding this part as part of the member name escaping PR.

packages/core/src/binder.ts
```diff
@@ -336,6 +336,12 @@ export interface Binder {
     name: string,
   ): Ref<TScope | undefined>;
 
+  /**
+   * Gets a symbol by its refkey.
+   * @param ref - The refkey to resolve.
+   */
+  getSymbolForRefkey(ref: Refkey): OutputSymbol | undefined;
```
bterlson, line 343: This needs to return a `Ref<OutputSymbol | undefined>`.

bterlson:
Closed in favor of #80.

PR #72, opened 2025-03-27
Add JSDoc support to TS Declarations

joheredi:
Add support for TS Declaration Docs


packages/typescript/src/components/DeclarationJSDoc.tsx
```diff
@@ -0,0 +1,18 @@
+import { Show } from "@alloy-js/core";
+import { JSDoc } from "./JSDoc.jsx";
+import { JSDocParagraph } from "./JSDocParagraph.jsx";
+
+export interface DeclarationJSDocProps {
+  doc?: string;
```
bterlson, line 6: I feel like this doc should support multiple paragraphs. Maybe take `string | string[]`?

packages/typescript/src/components/DeclarationJSDoc.tsx
```diff
@@ -0,0 +1,18 @@
...
+}
+
+export function DeclarationJSDoc(props: DeclarationJSDocProps) {
```
bterlson, line 9: This is what the existing `JSDoc` component is supposed to be. Seems the only difference is that this takes the doc string as a prop. You can add support for that to `JSDoc` instead I think?

packages/typescript/src/components/FunctionDeclarationJSDoc.tsx
```diff
@@ -0,0 +1,43 @@
...
+      props.parameters.length > 0,
+  );
+  const parameterDescriptors =
+    Array.isArray(props.parameters) ? props.parameters : [];
+  return (
+    <Show when={Boolean(props.doc)}>
+      <JSDoc>
+        <JSDocParagraph>{props.doc}</JSDocParagraph>
+        {canHaveParamDoc ?
+          <For each={parameterDescriptors}>
```
bterlson, line 26: Are we sure this component is necessary? It feels like it might be easier to put lines 26 -> 38 into a component like `<JSDocParameters>`, and then when creating docs for a function you use the same `JSDoc` component you would for anything else, except you also place a `<JSDocParameters>` inside.

```
<JSDoc>
  <JSDocParagraph>Cool function</JSDocParagraph>
  <JSDocParameters parameters={params} />
</JSDoc>
<FunctionDeclaration ... />
```

.chronus/changes/feature-ts-declaration-doc-2025-2-27-21-33-51.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: feature
+packages:
+  - "@alloy-js/typescript"
+---
+
+Add TS Declaration JSDoc
```
bterlson, line 7: Create changelog entry explicitly mentioning the components which were added.

bterlson (changes requested):

packages/typescript/src/components/ClassDeclaration.tsx
```diff
@@ -137,14 +153,27 @@ export interface ClassMethodProps extends ClassMemberProps {
 
 export function ClassMethod(props: ClassMethodProps) {
   const returnType = props.returnType && <>: {props.returnType}</>;
+  const [docProps, rest] = splitProps(props, ["doc"]);
```
bterlson, line 156: This splitProps seems unnecessary.

packages/typescript/src/components/Interface.tsx
```diff
@@ -60,6 +69,10 @@ export function InterfaceMember(props: InterfaceMemberProps) {
   } else {
     return (
       <>
+        <Show when={Boolean(props.doc)}>
+          <JSDoc children={props.doc} />
```
bterlson, line 73: Since you're just passing as children, you could make `doc` children as well, instead of just string. Or at least take an array of strings which will allow passing multiple paragraphs.

packages/typescript/src/components/ClassDeclaration.tsx
joheredi: I need to take out docs from props otherwise the memeber will try to render them again. Member also renders docs because a non method member needs simple JSDoc

PR #74, opened 2025-04-02
Implement call signature, clean up refkey, add defaultProps, clean up TS arguments

bterlson:
A batch of changes here:

* Implement a `CallSignature` component for TypeScript.
* As part of this, clean up the parameters and typeParameters types. NOTE: the record signatures are now gone completely.
* Add `defaultProps` to set defaults to props
* Remove plural `refkeys` anywhere it was added in lieu of just taking an array for the `refkey` prop.


packages/typescript/src/components/FunctionDeclaration.tsx
```diff
@@ -125,67 +162,47 @@ interface DeclaredParameterDescriptor
 }
 
 function normalizeAndDeclareParameters(
-  parameters: Record<string, Children> | ParameterDescriptor[] | string[],
+  parameters: ParameterDescriptor[] | string[],
   flags: TSSymbolFlags = TSSymbolFlags.ParameterSymbol,
 ): DeclaredParameterDescriptor[] {
   const namePolicy = useTSNamePolicy();
-  if (Array.isArray(parameters)) {
```
joheredi, line 132: Love that we can remove these kind of logic

packages/typescript/test/function-declaration.test.tsx
```diff
@@ -76,7 +76,10 @@ it("supports type parameters by descriptor object", () => {
   const decl = (
     <FunctionDeclaration
       name="foo"
-      typeParameters={{ a: "any", b: "any" }}
+      typeParameters={[
+        { name: "a", type: "any" },
```
joheredi, line 80: I think this should be `extends: "any"` instead of type?

joheredi (approved):

packages/typescript/test/function-declaration.test.tsx
bterlson: Right you are.

PR #91, opened 2025-04-09
Function and Arrow function expression components

### LINKED ISSUE #88: TypeScript needs Function Expression forms
Both normal function expressions and arrow functions.

timotheeguerin:
fix #88


packages/typescript/src/components/FunctionBase.tsx
timotheeguerin, line 1: Moved all the common building blocks for function here, not sure what is the current pattern we follow for those, not everything here needs to be rexported (selecting export in index.ts)

Should we move those building block to a sub folder or should the function components be its own folder with all the building block as siblings.

packages/typescript/src/components/FunctionArrowExpression.tsx
```diff
@@ -0,0 +1,92 @@
...
+  FunctionBody,
+  functionBodyTag,
+  FunctionParameters,
+  functionParametersTag,
+  FunctionTypeParameters,
+  functionTypeParametersTag,
+  getReturnType,
+} from "./FunctionBase.jsx";
+
+export interface FunctionArrowExpressionProps extends CallSignatureProps {
```
bterlson, line 21: ArrowFunctionExpression* I should think (ECMAScript grammar production)

bterlson: ES grammar doesn't have Expression actually, so fine leaving that out. Looks a lot better :)

bterlson (approved):

packages/typescript/src/components/FunctionBase.tsx
timotheeguerin: lets maybe see how other commponents evolve too and I'll refactor if needs be

PR #104, opened 2025-04-11
Add ReferenceOrContent Component

### LINKED ISSUE #90: Add core ReferenceOrContent component
This component should conditionally render its contents based on whether the refkey has an associated symbol. So for example:

```
const rk1 = refkey();
const rk2 = refkey();
return <List>
  <Declaration name="foo" refkey={rk1}>Declare foo</Declaration>
  <ReferenceOrContent refkey={rk1}>No Reference</ReferenceOrContent>
  <ReferenceOrContent refkey={rk2}>No Reference</ReferenceOrContent>
</List>
```

Should render (given an appropriate reference component):

```
Declare foo
foo
No Reference
```

### LINKED ISSUE #102: Alloy build --watch needs restart when adding new files
Probably because we resolve the files ahead of time, might be better to just default back to let the watch host load the config itself as well.

### LINKED ISSUE #103: Alloy --watch should clear screen

timotheeguerin:
fix #90 Add reference or context
fix #102 `alloy build --watch` handles new files being added or removed
fix #103 `alloy build --watch` clear screen


packages/core/src/components/ReferenceOrContent.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+export function ReferenceOrContent(props: ReferenceOrContentProps) {
+  const binder = useContext(BinderContext);
+  if (!binder) {
+    throw new Error("Need binder context to create declarations");
+  }
+
+  const sym = binder.getSymbolForRefkey(props.refkey);
+  return computed(() =>
+    sym.value === undefined ? props.children : sym.value.name,
```
timotheeguerin, line 20:
```suggestion
    sym.value === undefined ? props.children : sym.value,
```

PR #109, opened 2025-04-11
Return ref instead of symbol to resovle correctly imports and cors files

timotheeguerin:


packages/core/test/components/reference-or-content.test.tsx
```diff
@@ -16,12 +19,20 @@ function TestWrapper(props: { children: Children }) {
   const GetBinder = createTap(() => {
     return useBinder();
   });
+
+  function Reference(props: { refkey: Refkey }) {
```
bterlson, line 23: Assert this is called maybe?

bterlson (approved):

PR #112, opened 2025-04-14
Fix interface member references

### LINKED ISSUE #108: Interface members with refkeys cause issues in InterfaceExpression
InterfaceExpression doesn't create a member scope. Perhaps a similar pattern to object expressions should be used - object expressions will check if there is an assignment symbol and if so, wrap its children in a MemberScope with an owner of that assignment symbol. This would mean, given:

```ts
interface Foo {
  outer: {
    inner: string;
  }
}
```

that the outer symbol would become a `SymbolFlags.StaticMemberContainer` and the symbol for `inner` would be a static member of the `outer` symbol.

### LINKED ISSUE #111: Type reference to interface members seems broken
```ts
 <>
        <ts.InterfaceDeclaration name="Foo">
              <ts.InterfaceMember name="prop" type="string" refkey={ref} />
        </ts.InterfaceDeclaration>
        <line />
        <ts.InterfaceDeclaration name="Bar">
          <ts.InterfaceMember name="ref" type={<Reference refkey={ref} />} />
        </ts.InterfaceDeclaration>
      </>,
```

Produce `Foo.prop` but it should be `Foo["prop"]`

timotheeguerin:
fix #108 
fix #111

### DIFF
--- a/packages/typescript/src/components/Interface.tsx
+++ b/packages/typescript/src/components/Interface.tsx
@@ -1,7 +1,6 @@
 import {
   Block,
   Children,
-  MemberDeclaration,
   MemberScope,
   Name,
   OutputSymbolFlags,
@@ -15,8 +14,7 @@ import { useTSNamePolicy } from "../name-policy.js";
 import { BaseDeclarationProps, Declaration } from "./Declaration.js";
 import { JSDoc } from "./JSDoc.jsx";
 
-import { TSOutputScope } from "../symbols/scopes.js";
-import { createTSSymbol, TSSymbolFlags } from "../symbols/ts-output-symbol.js";
+import { MemberDeclaration } from "./MemberDeclaration.jsx";
 import { PropertyName } from "./PropertyName.jsx";
 
 export interface InterfaceDeclarationProps extends BaseDeclarationProps {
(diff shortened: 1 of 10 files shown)


packages/typescript/src/components/Interface.tsx
```diff
@@ -87,9 +108,18 @@ export function InterfaceMember(props: InterfaceMemberProps) {
...
+      const sym = createTSSymbol({
+        name,
+        scope: memberScope.staticMembers as TSOutputScope,
+        refkey: props.refkey,
+        flags: OutputSymbolFlags.StaticMember,
+        tsFlags: TSSymbolFlags.TypeSymbol,
+      });
       return (
-        <MemberDeclaration static name={name} refkey={props.refkey}>
+        <MemberDeclaration static symbol={sym} refkey={props.refkey}>
```
bterlson, line 122: don't need to pass refkey here now.

bterlson: Also could probably stand to have the TypeScript equivalent of `MemberDeclaration`.

bterlson (approved):
Small nits but this is amazing!

packages/typescript/src/components/MemberDeclaration.tsx
```diff
@@ -0,0 +1,70 @@
...
+   */
+  flags?: OutputSymbolFlags;
+
+  /**
+   * The symbol to use for this memberdeclaration.
+   */
+  symbol?: TSOutputSymbol;
+}
+
+export function MemberDeclaration(props: Readonly<MemberDeclarationProps>) {
```
bterlson, line 45: Can we get a changelog entry for this component?

bterlson (approved):

PR #114, opened 2025-04-16
Expand `instantiateSymbolInto` to fully mirror instance & static members

joheredi:
**What’s changed**  
- **Dual‑scope support**: No longer restrict to instance‑only. Now handles both `InstanceMemberContainer` and `StaticMemberContainer` flags.  
- **Scope prep**: Always calls `addInstanceMembersToSymbol` & `addStaticMembersToSymbol` so both target scopes exist.  
- **Unified recursion**: Introduces a single `copyMembers` helper that:  
  - Computes and reuses composite `refkey` to prevent duplicates (idempotency).  
  - Marks each clone with `StaticMember` for correct dot‑notation resolution.  
  - Recurses into nested `staticMemberScope` for deep static hierarchies.  
- **Error removal**: Removed the old guard that threw unless source was an instance container.  

**Testing**  
- Added/updated tests to cover:  
  - Initial and late addition of instance members  
  - Initial and late addition of static members  
  - Recursive nesting of static children  
  - Combined instance+static containers  
  - Idempotency on repeated calls.

### DIFF
--- a/packages/core/src/binder.ts
+++ b/packages/core/src/binder.ts
@@ -9,7 +9,7 @@ import {
 import { useBinder } from "./context/binder.js";
 import { useMemberScope } from "./context/member-scope.js";
 import { useScope } from "./context/scope.js";
-import { effect, memo, untrack } from "./jsx-runtime.js";
+import { effect, memo, onCleanup, untrack } from "./jsx-runtime.js";
 import { refkey, Refkey } from "./refkey.js";
 import { queueJob, QueueJob } from "./scheduler.js";
 export type Metadata = object;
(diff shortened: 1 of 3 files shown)


packages/core/src/binder.ts
```diff
@@ -622,26 +622,72 @@ export function createOutputBinder(options: BinderOptions = {}): Binder {
   }
 
   function instantiateSymbolInto(source: OutputSymbol, target: OutputSymbol) {
-    if (~source.flags & OutputSymbolFlags.InstanceMemberContainer) {
-      throw new Error("Can only instantiate symbols with instance members");
-    }
-
+    // Ensure both member scopes exist on the target
     addInstanceMembersToSymbol(target);
+    addStaticMembersToSymbol(target);
```
bterlson, line 627: Instantiation should only ever result in static member symbols. Instantiated member symbols on the source become static symbols on the target. I suspect you won't be able to reference instantiated member symbols with this implementation?

bterlson:
I believe `instantiateSymbolInto` is broken when source symbols are changed - it likely adds a whole new set of symbols. This should be confirmed with a test, and then fixed (I think by storing the added symbols and removing them in `onCleanup`).

PR #117, opened 2025-04-16
Add nullish to TS Symbol flags

joheredi:
* Adds a nullish flag to TSSymbolFlag
* Automatically tag optional class and interface members as nullish
* Update reference member expression to handle nullish
* Update MemberExpression component to handle nullish

Todo:
* Explore if we need automatic flagging for function parameters

### DIFF
--- a/packages/typescript/src/components/ClassDeclaration.tsx
+++ b/packages/typescript/src/components/ClassDeclaration.tsx
@@ -22,6 +22,7 @@ import { BaseDeclarationProps, Declaration } from "./Declaration.jsx";
 import { JSDoc } from "./JSDoc.jsx";
 import { JSDocParams } from "./JSDocParam.jsx";
 import { PropertyName } from "./PropertyName.jsx";
+import { TypeRefContext } from "./TypeRefContext.jsx";
 
 export interface ClassDeclarationProps extends BaseDeclarationProps {
   extends?: Children;
--- a/packages/typescript/src/components/ClassDeclaration.tsx
+++ b/packages/typescript/src/components/ClassDeclaration.tsx
@@ -101,6 +101,7 @@ export interface ClassMemberProps {
   static?: boolean;
   children?: Children;
   doc?: Children;
+  nullish?: boolean;
 }
 
 export function ClassMember(props: ClassMemberProps) {
@@ -128,12 +129,19 @@ export function ClassMember(props: ClassMemberProps) {
       : memberScope.instanceMembers!) as TSOutputScope;
   }
 
+  let tsFlags =
+    props.jsPrivate ? TSSymbolFlags.PrivateMember : TSSymbolFlags.None;
+
+  if (props.nullish) {
+    tsFlags |= TSSymbolFlags.Nullish;
+  }
+
   const sym = createTSSymbol({
     name,
     scope,
     refkey: props.refkey,
     flags,
-    tsFlags: props.jsPrivate ? TSSymbolFlags.PrivateMember : TSSymbolFlags.None,
+    tsFlags,
   });
 
   return (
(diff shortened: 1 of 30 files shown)


packages/typescript/src/symbols/reference.ts
```diff
@@ -131,17 +135,21 @@ function buildMemberExpression(path: TSOutputSymbol[]) {
     path = path.slice(1);
   }
 
+  let nullish = isNullish(base);
```
bterlson, line 138:
```suggestion
  let prevNullish = isNullish(base);
```

packages/typescript/src/symbols/ts-output-symbol.ts
```diff
@@ -78,3 +79,7 @@ export function createTSSymbol(options: createTsSymbolOptions): TSOutputSymbol {
 
   return sym;
 }
+
+export function isNullish(sym: OutputSymbol): boolean {
```
bterlson, line 83: I generally don't find this kind of helper useful, the bitwise op is pretty terse. No real problem with it if we can avoid making it public API though.

bterlson (changes requested):
This is good. I think we should include support for nullish function parameters as part of this. The descriptor should probably allow specifying nullish, and optional parameters are nullish by default.

packages/typescript/src/components/Interface.tsx
```diff
@@ -116,6 +116,7 @@ export function InterfaceMember(props: InterfaceMemberProps) {
           kind="type"
           nameKind="interface-member"
           flags={OutputSymbolFlags.StaticMember}
+          nullish={props.optional}
```
timotheeguerin, line 119: I assume the use case here is its up to the user to know that the type can have `null` as a variant and so we pass this to be explicit?

packages/typescript/src/symbols/ts-output-symbol.ts
joheredi: Yeah tat's fair

packages/typescript/src/components/Interface.tsx
joheredi: Yeah, and thinking on this, maybe we should add 'nullish' prop to InterfaceMemberProps as well

packages/typescript/src/components/ClassDeclaration.tsx
```diff
@@ -155,15 +163,22 @@ export function ClassMember(props: ClassMemberProps) {
...
+  optional?: boolean;
   children?: Children;
 }
 
 export function ClassField(props: ClassFieldProps) {
-  const typeSection = props.type && <>: {props.type}</>;
+  const optionality = props.optional ? "?" : "";
+  const typeSection = props.type && (
+    <>
+      {optionality}: {props.type}
```
bterlson, line 174: Needs the new `<TypeRefContext>` wrapper.

bterlson, line 174:
```suggestion
      {optionality}: <TypeRefContext>{props.type}</TypeRefContext>
```

packages/typescript/test/class.test.tsx
```diff
@@ -358,3 +358,22 @@ it("renders a method with parameter docs", () => {
     }
   `);
 });
+
+it("creates a nullish symbol for an optional member", () => {
```
bterlson, line 362: This test doesn't actually validate that a nullish symbol is created.

packages/typescript/test/member-expression.test.tsx
```diff
@@ -143,6 +153,164 @@ describe("with refkeys", () => {
     `);
   });
 
+  it("handles nullish symbols correctly", () => {
```
bterlson, line 156: Might be good to note that this test is deliberately producing invalid references (it's just easiest to test with VarDeclaration).

packages/typescript/test/class.test.tsx
joheredi: Oh yeah moved these to MemberExpression

PR #124, opened 2025-04-18
Type references

### LINKED ISSUE #120: Support typeOnly references.
There should be a way to reference just a type such that any imports generated include the `type` modifier. It may be possible to do this automatically in type contexts, but there should still probably be a low-level way to do it e.g. via `<Reference refkey={} type />`.

timotheeguerin:
fix #120

when using a typescript reference can specify it is used in a type context which updates the imports to include `type` keyword. 

If a symbol is used both as type and non type the keyword will be omitted.
If all symbols from an import are types it uses this syntax `import type {A, B, C} from "..."` otherwise it use `import {type A, B, type C} from "..."`

### DIFF
--- a/packages/typescript/src/components/Reference.tsx
+++ b/packages/typescript/src/components/Reference.tsx
@@ -3,10 +3,16 @@ import { ref } from "../symbols/index.js";
 
 export interface ReferenceProps {
   refkey: Refkey;
+
+  /**
+   * If the reference is a type.
+   * This affect things like import where `type` keyword might be used.
+   */
+  type?: boolean;
 }
 
-export function Reference({ refkey }: ReferenceProps) {
-  const reference = ref(refkey);
+export function Reference({ refkey, type }: Readonly<ReferenceProps>) {
+  const reference = ref(refkey, type);
 
   return <>{reference}</>;
 }
--- a/packages/typescript/src/components/Reference.tsx
+++ b/packages/typescript/src/components/Reference.tsx
@@ -6,13 +6,15 @@ export interface ReferenceProps {
 
   /**
    * Whether this is a reference to a type.
+   *
+   * @remarks
    * This affect things like import where `type` keyword might be used.
    */
   type?: boolean;
 }
 
 export function Reference({ refkey, type }: Readonly<ReferenceProps>) {
-  const reference = ref(refkey, type);
+  const reference = ref(refkey, { type });
 
   return <>{reference}</>;
 }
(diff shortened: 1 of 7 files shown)


packages/typescript/src/symbols/ts-module-scope.ts
```diff
@@ -19,7 +19,11 @@ export interface TSModuleScope extends OutputScope {
    */
   importedSymbols: Map<TSOutputSymbol, TSOutputSymbol>;
   importedModules: ImportRecords;
-  addImport(symbol: TSOutputSymbol, module: TSModuleScope): TSOutputSymbol;
+  addImport(
+    symbol: TSOutputSymbol,
+    module: TSModuleScope,
+    type?: boolean,
```
timotheeguerin, line 25: Should maybe have an option bag instead of boolean positional.

It didn't feel right to have this be the symbol flag as they don't really apply here.

packages/typescript/src/components/Reference.tsx
```diff
@@ -3,10 +3,16 @@ import { ref } from "../symbols/index.js";
 
 export interface ReferenceProps {
   refkey: Refkey;
+
+  /**
+   * If the reference is a type.
```
bterlson, line 8:
```suggestion
   * Wether this is a reference to a type.
```

packages/typescript/src/components/Reference.tsx
```diff
@@ -3,10 +3,16 @@ import { ref } from "../symbols/index.js";
...
+   * This affect things like import where `type` keyword might be used.
```
bterlson, line 9: Put under @remarks probably?

packages/typescript/src/symbols/reference.ts
```diff
@@ -20,7 +20,7 @@ import { TSModuleScope } from "./ts-module-scope.js";
 import { TSOutputSymbol, TSSymbolFlags } from "./ts-output-symbol.js";
 import { TSPackageScope } from "./ts-package-scope.js";
 
-export function ref(refkey: Refkey): () => string {
+export function ref(refkey: Refkey, type?: boolean): () => string {
```
bterlson, line 23: Options bag here would be nice (can imagine other reference customizations that might need to be done, e.g. the selectors tsdoc supports to disambiguate the various parts of a type).

packages/typescript/src/symbols/ts-module-scope.ts
bterlson: Yeah I think options bag here too (though I care less since this is pretty much internal)

bterlson (approved):

PR #125, opened 2025-04-18
Type context

timotheeguerin:
Follow up to #120 and #124 which automatically define the current context (value vs type)

Automatically enter this context when in
- interface
- interface expression
- function type


packages/typescript/src/context/ts-context.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export interface TypeScriptContext {
+  /** If in a type context */
+  type?: boolean;
+}
+
+/**
+ * Provides scopes for instance and static private members.
+ */
+export const TypeScriptContext: ComponentContext<TypeScriptContext> =
```
timotheeguerin, line 11: Not really sure about the name for this or even if we should reuse something like scope(Didn't felt right)

packages/typescript/src/context/ts-context.ts
```diff
@@ -0,0 +1,16 @@
+import { ComponentContext, createContext, useContext } from "@alloy-js/core";
+
+export interface TypeScriptContext {
+  /** If in a type context */
+  type?: boolean;
```
timotheeguerin, line 5: Would it be better to have `value | type`?

packages/typescript/src/components/FunctionType.tsx
```diff
@@ -65,7 +66,7 @@ export function FunctionType(props: FunctionTypeProps) {
   });
 
   return (
-    <>
+    <TypeScriptContext.Provider value={{ type: true }}>
```
timotheeguerin, line 69: Might be nicer to have a shorter name

packages/typescript/src/context/ts-context.ts
```diff
@@ -0,0 +1,16 @@
...
+}
+
+/**
+ * Provides scopes for instance and static private members.
+ */
+export const TypeScriptContext: ComponentContext<TypeScriptContext> =
+  createContext<TypeScriptContext>();
+
+export function useTSContext(): TypeScriptContext {
+  return useContext(TypeScriptContext) ?? { type: false };
```
bterlson, line 15: You can set the default as the first argument to createContext.

packages/typescript/src/context/ts-context.ts
bterlson: I don't think we want anything generalized here. I'd do `TypeRefContext: ComponentContext<TypeRefContext>` maybe?

I would also add `isTypeRefContext()` accessor.

bterlson (changes requested):
This is a super exciting direction.

I think my only overall concern is where we should actually set this context. I feel like there are at least three reasonable(?) options:

* On every type declaration and expression component. This has the effect of polluting the tree with a ton of useless context nodes.
* Only on boundary nodes with type-land, so TypeDeclaration and the other *Declaration components which produce types, any component which creates a value and takes a `type` prop (var declaration, interface prop, parameter, etc.). Easy to miss a place, conceivable to end up in TypeExpression land without the context set which would be a sad day.
* All components which define types check to see if they're in type land and if not, set the context. This is a lot of boilerplate.

packages/typescript/src/context/ts-context.ts
timotheeguerin: created a new component that hides that, let me know what you think

bterlson (approved):

PR #127, opened 2025-04-20
Use Scheduler

bterlson:
This introduces a scheduler to handle invoking reactive effects. For now it is merely a queue, which has the effect of making effects FIFO instead of LIFO. Additionally, since the queue is a set keyed by the effect function, we end up deduplicating effects. Consider the following (assuming we're in a component or other code run during rendering):

```js
let obj = reactive({ a: 0, a: 0 });

effect(() => {
  console.log({ a: obj.a, b: obj.b });
});
console.log("1");
obj.a = 1;
console.log("2");
obj.b = 2;
console.log("3");
```

Previously this would log `1, { a: 1, b: 0 }, 2, { a: 1, b: 2 }, 3`, but now logs `1, 2, 3, { a: 1, b: 2}`.

This coalescing effect seems to be a fairly significant optimization. Symbols especially frequently triggered multiple duplicate effects doing a lot of pointless work. Additionally, the scheduler provides an easy way to avoid duplicate work not tied to effects. This is used to make name conflict resolution run once all symbols have been declared in the current effect (so e.g. parameter conflicts will be called once per parameter list, not once per conflicted symbol). With a priority queue this could be updated to be called essentially once per scope by making the conflict resolution low priority.

This change is technically breaking, as if you are changing reactive values/signals after calling render/renderTree/printTree, you will need to call `flushJobs` yourself. I suspect this is only limited to tests (see the various test updates in this PR), and likely only tests in the core framework, so hopefully this breaking change is minor.

Another consequence of this change is that it is somewhat less subtle avoiding infinite loops with recursive effects. This PR allows recursive effects in order to allow patterns like dynamic discovery of types to emit while emitting types (see new test). However, for complex reactive code, you can definitely still get into recursive/mutually recursive effects without trying too hard. See for example the fix to the binder. Generally the fix is to slap an untrack somewhere. I think this is ok under the philosophy that complex reactive code will be mostly limited to the framework. In the future we can consider allowing recursive effects in some places but not others (e.g. in Vue, render effects are allowed to be recursive but manual effects are not).

### DIFF
--- a/.chronus/changes/scheduler-2025-3-20-16-9-33.md
+++ b/.chronus/changes/scheduler-2025-3-20-16-9-33.md
@@ -1,0 +1,7 @@
+---
+changeKind: breaking
+packages:
+  - "@alloy-js/core"
+---
+
+Use a scheduler for effects. Effects are no longer run instantly but are queued to run after the current effect finishes. This significantly reduces effect calls due to coalescing multiple updates and also enables more recursive patterns.
--- a/packages/core/src/binder.ts
+++ b/packages/core/src/binder.ts
@@ -1,6 +1,5 @@
 import {
   computed,
-  effect,
   reactive,
   ref,
   Ref,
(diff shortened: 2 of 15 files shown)


bterlson:
Validated with TypeSpec JS Client emitter, no regressions, very slight perf improvement (test times dominated by tsp compile)

timotheeguerin (approved):

PR #130, opened 2025-04-22
Add watch timestamp

timotheeguerin:
image

### DIFF
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -112,6 +112,11 @@ function reportDiagnostics(diagnostics: readonly ts.Diagnostic[]) {
  * This is mainly for messages like "Starting compilation" or "Compilation completed".
  */
 function reportWatchStatusChanged(diagnostic: ts.Diagnostic) {
+  const time = new Date();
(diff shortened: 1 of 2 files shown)


packages/cli/src/cli.ts
```diff
@@ -112,6 +112,11 @@ function reportDiagnostics(diagnostics: readonly ts.Diagnostic[]) {
  * This is mainly for messages like "Starting compilation" or "Compilation completed".
  */
 function reportWatchStatusChanged(diagnostic: ts.Diagnostic) {
+  const time = new Date();
+  const hours = time.getHours();
+  const amPm = hours >= 12 ? "PM" : "AM";
+  const hours12 = hours % 12 || 12;
+  const formattedTime = `${hours12}:${time.getMinutes()}:${time.getSeconds()} ${amPm}`;
```
maorleger, line 119: You might use `toLocaleTimeString` like
```ts
  const formattedTime = time.toLocaleTimeString("en-US", {
    hour: "numeric",
    minute: "2-digit",
    second: "2-digit",
    hour12: true,
  });
```

Which I think does the same thing

PR #145, opened 2025-05-02
Sort import statements and named imports

timotheeguerin:
1. Sorting the statements by path
2. Sorting the named import by local path(Seems to be the same as prettier does though by target path would maybe make more sense)


packages/typescript/src/components/ImportStatement.tsx
```diff
@@ -21,9 +21,14 @@ export interface ImportStatementsProps {
 export function ImportStatements(props: ImportStatementsProps) {
   const pkg = usePackage();
 
+  const imports = [...props.records];
```
bterlson, line 24: I'm surprised this works given that this set is not reactive. I think that suggests that this entire component is getting re-rendered whenever an import is added?

PR #146, opened 2025-05-02
Rest and spread

timotheeguerin:


packages/typescript/src/components/ObjectExpression.tsx
```diff
@@ -138,3 +138,23 @@ export function ObjectProperty(props: ObjectPropertyProps) {
...
+  value?: Children;
+  jsValue?: unknown;
+  children?: Children;
+}
+
+export function ObjectSpreadProperty(props: ObjectSpreadPropertyProps) {
+  let value;
+  if (props.value) {
+    value = props.value;
+  } else if (Object.prototype.hasOwnProperty.call(props, "jsValue")) {
```
bterlson, line 152: Can use Object.hasOwn for this I think 😁

timotheeguerin: ah just copied the code above, but can change both

PR #152, opened 2025-05-05
Markdown: Add Code component

timotheeguerin:


packages/markdown/src/components/Code.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * Render a Markdown code block
+ */
+export function Code(props: CodeProps) {
+  const tripleBackticks = "```";
+  return (
+    <List>
+      {tripleBackticks + props.lang}
+      {props.content ?? <List children={props.children} />}
```
bterlson, line 22: I think we don't want the List for the `Code` component because we don't want to mess up formatting of the code contained therein. I think drop the content prop and delegate all formatting to the Children.

bterlson (changes requested):

packages/markdown/src/components/Code.tsx
```diff
@@ -0,0 +1,26 @@
+import { List } from "@alloy-js/core";
+import { Children } from "@alloy-js/core/jsx-runtime";
+
+export interface CodeProps {
+  /** Language of the code block */
+  lang?: string;
+
+  /** Content of the codeblock, use instead of children. */
```
bterlson, line 8:
```suggestion
  /** Content of the code block */
```

Was a little confusing because it feels like its saying, don't use children...

That said, this does the same thing as setting children to the string you pass to content, so I'm not sure this prop is worth having? I could see the argument though. If you like it, fix up the docstring and merge!

PR #153, opened 2025-05-05
createPackage should allow defining static and instance members

### LINKED ISSUE #87: createPackage should allow defining static members
E.g. I should be able to do something like:

```ts

export const mcpSdk = createPackage({
  name: "@modelcontextprotocol/sdk",
  version: "^3.23.0",
  descriptor: {
    "./server/index.js": {
      named: [
        { name: "server", staticMembers: ["setRequestHandler"] } // descriptor form
        "other" // simple form
      ],
    },
  }
})'

mcpSdk["./server/index.js"].server.setRequestHandler;
```

maorleger:
This pull request introduces a new feature to the `@alloy-js/typescript` package, enabling support for named static and instance members in module descriptors. The changes include updates to the `createPackage` API, new utility types, and additional test cases to ensure the functionality works as intended.

This supports scenarios such as:

```typescript
  const mcpSdk = createPackage({
    name: "@modelcontextprotocol/sdk",
    version: "^3.23.0",
    descriptor: {
      "./server/index.js": {
        named: [
          {
            name: "server",
            staticMembers: [
              "setRequestHandler",
              { name: "nested", staticMembers: ["nestedHandler"] },
            ],
            instanceMembers: ["instanceHandler"],
          },
          { name: "noMembers" },
          "simpleName",
        ],
      },
    },
  });
```

With namespacing and type inference providing the correct mappings.

Resolves #87

### DIFF
--- a/.chronus/changes/static-members-2025-4-7-20-40-48.md
+++ b/.chronus/changes/static-members-2025-4-7-20-40-48.md
@@ -4,4 +4,4 @@ packages:
   - "@alloy-js/typescript"
 ---
 
-Add support to createPackage for providing named members
+Add support to createPackage for defining instance and static members.
--- a/.chronus/changes/static-members-2025-4-7-20-40-48.md
+++ b/.chronus/changes/static-members-2025-4-7-20-40-48.md
@@ -1,0 +1,7 @@
+---
+changeKind: feature
+packages:
+  - "@alloy-js/typescript"
+---
+
+Add support to createPackage for providing named members
--- a/packages/typescript/src/create-package.ts
+++ b/packages/typescript/src/create-package.ts
@@ -65,7 +65,6 @@ function createStaticMembers(
   for (const member of staticMembers) {
     const memberObj = typeof member === "object" ? member : { name: member };
 
-    console.log("creating static member", memberObj);
     // Create a refkey directly if it doesn't exist yet
     keys[memberObj.name] ??= refkey();
 
@@ -87,11 +86,7 @@ function createStaticMembers(
       default: false,
       flags: memberFlags,
     });
-    memberScope.symbols.forEach((s) => {
-      console.log("member scope", s.name);
-    });
     memberScope.symbols.add(memberSym);
-    console.log("created static member", memberSym);
 
     // Handle static members of the current member
     createStaticMembers(
@@ -124,7 +119,6 @@ function createInstanceMembers(
 
   for (const member of instanceMembers) {
     const memberObj = typeof member === "object" ? member : { name: member };
-    console.log("creating instance member", memberObj);
 
     // Create a refkey directly if it doesn't exist yet
     if (!keys[memberObj.name]) {
@@ -149,7 +143,6 @@ function createInstanceMembers(
     });
 
     ownerSym.instanceMemberScope.symbols.add(memberSym);
-    console.log("created instance member", memberSym);
 
     // Recursively handle nested static members
     createStaticMembers(
@@ -361,6 +354,5 @@ export function createPackage<const T extends PackageDescriptor>(
     }
   }
 
-  console.log("all my refkeys", refkeys);
   return refkeys;
 }
--- a/packages/typescript/src/create-package.ts
+++ b/packages/typescript/src/create-package.ts
@@ -206,6 +206,26 @@ function createSymbols(
   }
 }
 
+function createRefkeysForMembers(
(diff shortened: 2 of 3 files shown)


packages/typescript/src/create-package.ts
```diff
@@ -54,28 +58,81 @@ function createSymbols(
...
-      const key = keys[exportedName];
-      const sym = createTSSymbol({
-        name: exportedName,
-        scope: moduleScope,
-        refkey: key,
-        export: true,
-        default: false,
-      });
-      moduleScope.exportedSymbols.set(key, sym);
+      if (typeof exportedName === "object") {
```
maorleger, line 61: purposely split this without refactoring to show the new code path for discussion

packages/typescript/src/create-package.ts
```diff
@@ -54,28 +58,81 @@ function createSymbols(
...
+        });
+        moduleScope.exportedSymbols.set(key, sym);
+      }
     }
   }
 }
 
+// TODO: simplify this crazy type?
+// Helper type to extract named exports as strings
+type NamedStringExports<T extends ModuleSymbolsDescriptor> = {
```
maorleger, line 101: Madness 😄

packages/typescript/src/create-package.ts
```diff
@@ -16,9 +16,13 @@ export interface PackageDescriptor {
   [path: string]: ModuleSymbolsDescriptor;
 }
 
+export type NamedModuleDescriptor =
+  | string
+  | { name: string; staticMembers?: Array<string> }; // TODO: array of strings sufficient?
```
bterlson, line 21: It is not, because static members may have their own static members, e.g. `someLib.foo.bar`.

packages/typescript/src/create-package.ts
```diff
@@ -54,28 +58,81 @@ function createSymbols(
...
+          name,
+          scope: moduleScope,
+          refkey: key,
+          export: true,
+          default: false,
+        });
+        moduleScope.exportedSymbols.set(key, sym);
+
+        for (const staticMember of staticMembers ?? []) {
+          const staticKey = refkey(`${key}.${staticMember}`);
```
bterlson, line 74: I think this refkey can just be a fresh refkey (no args).

packages/typescript/src/create-package.ts
bterlson: I mean, if it works....

bterlson:
I think the signature we want is going to look somewhat similar to the helper in test/symbols.test.ts named `createScopeTree`. You'll have a top-level package descriptor, which has exported symbol descriptors, each of which has instance and static member members which take more symbol descriptors. So you might have something like:

```
type SymbolDescriptor = string | { name: string, flags?: string, instanceMembers?: SymbolDescriptor[], staticMembers?: SymbolDescriptor[] };
```

When instanceMembers is provided you can automatically add SymbolFlags.InstanceMemberContainer to the containing symbol and stuff like that.

packages/typescript/src/create-package.ts
bterlson: Let's see... The package defines an exported module "./server/index.js" with a named export of "server" (`SymbolFlags.StaticMemberContainer`), which has a static member symbol "nested" (`SymbolFlags.StaticMember | SymbolFlags.StaticMemberContainer`), which has a static member symbol "nestedHandler" (`SymbolFlags.StaticMember`)

Then yes I would expect `mcpSdk["./server/index.js"].server.nested.nestedHandler` to vend a refkey for the `nestedHandler` symbol, and when resolving that refkey, to get the correct reference to it (which should hopefully work without any changes to referencing code which should already support this logic).

packages/typescript/src/create-package.ts
maorleger: Fair enough. I tried to refactor it a bit where possible

maorleger:
Nothing to see here 🙈 

Got some cleaning up and refactoring to do, then I will mark it ready for review

bterlson (approved):

.chronus/changes/static-members-2025-4-7-20-40-48.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: feature
+packages:
+  - "@alloy-js/typescript"
+---
+
+Add support to createPackage for providing named members
```
bterlson, line 7:
```suggestion
Add support to createPackage for defining instance and static members.
```

PR #155, opened 2025-05-08
Add a json based livr schema emitter

swatkatz:
We add a new sample program to generate a LIVR compatible JSON schema using the json-alloy framework. This is to add to the existing samples.

The code works by running: `node dist/src/index.js` assuming that the you are running `pnpm watch` inside of `alloy/samples/livr-emitter`

### DIFF
--- a/samples/livr-emitter/src/schema.ts
+++ b/samples/livr-emitter/src/schema.ts
@@ -1,0 +1,290 @@
+/**
+ * Base interface for all LIVR rules.
+ * Extend and discriminate with `rule` property.
+ */
+interface LIVRRuleBase {
+  rule: string;
+  error_code?: string;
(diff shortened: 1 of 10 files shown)


samples/livr-emitter/src/schema.ts
```diff
@@ -0,0 +1,290 @@
...
+ */
+interface LIVRRuleBase {
+  rule: string;
+  error_code?: string;
+}
+
+// --- Common rules ---
+export interface RequiredRule extends LIVRRuleBase {
+  rule: "required";
+  error_code?: "REQUIRED";
```
steverice, line 13: Should we create an enum to represent the error codes? Many of them are used in multiple places.

PR #156, opened 2025-05-12
Add initial python bindings

swatkatz:
This commit adds some basic python bindings for 
1. SourceFile
2. Imports
3. PythonSymbol

These were required to test the imports basically. They are not fully done yet, and will probably undergo changes as more components are created.

These were added to make sure that tests pass.

### DIFF
--- a/packages/python/src/components/ImportStatement.tsx
+++ b/packages/python/src/components/ImportStatement.tsx
@@ -1,0 +1,57 @@
+import { computed, JSX, mapJoin, memo } from "@alloy-js/core";
+
+export interface ImportSymbol {
+  module: string; // The module to import from
+  names?: Array<string | { name: string; alias?: string }>; // Items to import
+  alias?: string; // Alias for the module itself (if importing the whole module)
+  wildcard?: boolean; // If true, use '*'
+}
+
+export interface ImportStatementsProps {
+  imports: ImportSymbol[];
+}
+
+/**
+ * Represents a Python import statement.
+ * Generates an import statement for the given module and names.
+ */
+export function ImportStatements(props: ImportStatementsProps): JSX.Element {
+  const imports = computed(() =>
+    props.imports.sort((a, b) => a.module.localeCompare(b.module)),
+  );
+
+  return mapJoin(
+    () => imports.value,
+    (importProp) => <ImportStatement {...importProp} />,
+  );
+}
+
+export function ImportStatement(props: ImportSymbol) {
+  return memo(() => {
+    const { module, names, alias, wildcard } = props;
+
+    const parts: any[] = [];
+
+    if (wildcard) {
+      parts.push(`from ${module} import *`);
+    } else if (!names || names.length === 0) {
+      parts.push(`import ${module}`);
+    } else {
+      const formattedNames = names
+        .map((name) =>
+          typeof name === "string" ? name
+          : name.alias ? `${name.name} as ${name.alias}`
+          : name.name,
+        )
+        .join(", ");
+
+      if (alias) {
+        parts.push(`from ${module} import ${formattedNames} as ${alias}`);
+      } else {
+        parts.push(`from ${module} import ${formattedNames}`);
+      }
+    }
+
+    return parts;
+  });
+}
--- a/packages/python/src/symbols/python-output-symbol.ts
+++ b/packages/python/src/symbols/python-output-symbol.ts
@@ -1,0 +1,38 @@
+import {
+  Children,
+  OutputSymbol,
+  refkey,
+  Refkey,
+  useBinder,
+  useScope,
+} from "@alloy-js/core";
+
+/**
+ * Represents an 'exported' symbol from a .py file. Class, enum, interface etc.
+ */
+export interface PythonOutputSymbol extends OutputSymbol {
+  // The module in which the symbol is defined
+  module?: string;
+}
+
+export interface createPythonSymbolProps {
+  name: string;
+  module?: string;
+  refKey?: Refkey | Refkey[];
+  children?: Children;
+}
+
+export function createPythonSymbol(
+  props: createPythonSymbolProps,
+): PythonOutputSymbol {
+  const binder = useBinder();
+  const scope = useScope();
+
+  const sym = binder.createSymbol<PythonOutputSymbol>({
+    name: props.name,
+    scope,
+    refkey: props.refKey ?? refkey(props.name),
+    module: props.module ? props.module : "", //TODO: Write a ModuleComponent to handle this
+  });
+  return sym;
+}
--- a/packages/python/test/vitest.d.ts
+++ b/packages/python/test/vitest.d.ts
@@ -1,0 +1,10 @@
+import "vitest";
+
+interface CustomMatchers<R = unknown> {
+  toRenderTo: (str: string) => R;
+}
+
+declare module "vitest" {
+  interface Assertion<T = any> extends CustomMatchers<T> {}
+  interface AsymmetricMatchersContaining extends CustomMatchers {}
+}
--- a/packages/python/CHANGELOG.md
+++ b/packages/python/CHANGELOG.md
@@ -1,0 +1,3 @@
+# Changelog - @alloy-js/python
+
+This package is currently in development. No releases yet.
--- a/packages/python/api-extractor.json
+++ b/packages/python/api-extractor.json
@@ -1,0 +1,4 @@
+{
+  "$schema": "https://developer.microsoft.com/json-schemas/api-extractor/v7/api-extractor.schema.json",
+  "extends": "../../api-extractor.base.json"
+}
--- a/packages/python/package.json
+++ b/packages/python/package.json
@@ -1,0 +1,42 @@
+{
+  "name": "@alloy-js/python",
+  "version": "0.0.1",
+  "description": "Python bindings for Alloy",
+  "exports": {
+    ".": {
+      "development": "./src/index.ts",
+      "import": "./dist/src/index.js"
+    },
+    "./stc": {
+      "development": "./src/components/stc/index.ts",
+      "import": "./dist/src/components/stc/index.js"
+    }
+  },
+  "scripts": {
+    "generate-docs": "api-extractor run",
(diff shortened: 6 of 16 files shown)


packages/python/src/symbols/python-output-symbol.ts
```diff
@@ -0,0 +1,38 @@
...
+  refkey,
+  Refkey,
+  useBinder,
+  useScope,
+} from "@alloy-js/core";
+
+/**
+ * Represents an 'exported' symbol from a .py file. Class, enum, interface etc.
+ */
+export interface PythonOutputSymbol extends OutputSymbol {
```
bterlson, line 13: Seems duplicated from the python-output-symbol.ts

swatkatz: sorry I am ts noob, so I might be totally misunderstanding your comment: I thought I've only defined this interface in `python-output-symbol.ts` which is this file?

packages/python/src/components/Class.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+export interface ClassProps extends DeclarationProps {
+  name: string;
+  bases?: Children[];
+}
+
+export function Class(props: ClassProps) {
+  const name = usePythonNamePolicy().getName(props.name, "class");
+  // Determine if children are present
+  const hasChildren =
```
bterlson, line 13: use childrenArray to get the children as an array.

packages/python/src/components/ImportStatement.tsx
```diff
@@ -0,0 +1,57 @@
+import { computed, JSX, mapJoin, memo } from "@alloy-js/core";
+
+export interface ImportSymbol {
+  module: string; // The module to import from
```
bterlson, line 4: Want to use jsdoc comments so the documentation system will pick it up.

packages/python/src/components/Method.tsx
```diff
@@ -0,0 +1,57 @@
+import { Children, Indent, Scope } from "@alloy-js/core";
+import { usePythonNamePolicy } from "../name-policy.js";
+import { Declaration } from "./Declaration.jsx";
+import { Parameters, ParametersProps } from "./Parameters.jsx";
+
+export interface MethodProps extends ParametersProps {
+  name: string; // e.g. "__init__" or "foo"
+  isInstanceMethod?: boolean; // true if this is an instance method
```
bterlson, line 8: I would call these `instance` and `class`, or `instanceMethod` and `classMethod`. Is prefix typically isn't used in the context of JSX component boolean props.

packages/python/src/components/Value.tsx
```diff
@@ -0,0 +1,62 @@
+import { For, Indent, memo } from "@alloy-js/core";
+
+export interface ValueProps {
+  value?: unknown;
+}
+
+export function Value(props: ValueProps): any {
```
bterlson, line 7: We typically want to stick to the python names for things, this appears to be an Atom, consider naming it that?

packages/python/src/components/Value.tsx
```diff
@@ -0,0 +1,62 @@
+import { For, Indent, memo } from "@alloy-js/core";
+
+export interface ValueProps {
+  value?: unknown;
```
bterlson, line 4: `jsValue` is the convention used when providing a concrete value in JSON that is serialized into the target language.

packages/python/src/components/Variable.tsx
```diff
@@ -0,0 +1,24 @@
+import { Children, code } from "@alloy-js/core";
+import { usePythonNamePolicy } from "../name-policy.js"; // assuming you have this
+
+export interface VariableProps {
+  name: string;
+  value?: Children;
+  type?: Children; // Optional, only for type annotation
+  omitNone?: boolean; // Optional, to omit None assignment
```
bterlson, line 8: This configuration seems a bit strange. Is it the case that if I provide a type, that none may not be the best default? I feel like it's ok to let initialization be up to the caller.

packages/python/src/components/Variable.tsx
```diff
@@ -0,0 +1,24 @@
+import { Children, code } from "@alloy-js/core";
+import { usePythonNamePolicy } from "../name-policy.js"; // assuming you have this
+
+export interface VariableProps {
```
bterlson, line 4: I think this should also take children like TS VarDeclaration. When you are providing complex initializers it is quite annoying to pass that initializer in as a property as opposed to just placing it in the children of Variable.

packages/python/src/components/Variable.tsx
```diff
@@ -0,0 +1,24 @@
...
+  name: string;
+  value?: Children;
+  type?: Children; // Optional, only for type annotation
+  omitNone?: boolean; // Optional, to omit None assignment
+}
+
+export function Variable(props: VariableProps) {
```
bterlson, line 11: AssignmentStatement?

bterlson: The python grammar is AssignmentStatement but I think it's more valuable to have `Declaration` in the name of components which make declarations, and `AssignmentDeclaration` seems weird to me, but I also wouldn't be opposed to that.

packages/python/src/components/Class.tsx
```diff
@@ -0,0 +1,30 @@
+import { Children, Indent, List, Scope, Show } from "@alloy-js/core";
+import { usePythonNamePolicy } from "../name-policy.js";
+import { Declaration, DeclarationProps } from "./Declaration.js";
+
+export interface ClassProps extends DeclarationProps {
+  name: string;
+  bases?: Children[];
+}
+
+export function Class(props: ClassProps) {
```
bterlson, line 10: ClassDeclaration

packages/python/src/components/Enum.tsx
```diff
@@ -0,0 +1,98 @@
...
+  /**
+   * Optional refkey for symbol referencing.
+   */
+  refkey?: Refkey;
+}
+
+/**
+ * A Python enum declaration, following https://docs.python.org/3.11/library/enum.html
+ */
+export function Enum(props: EnumProps) {
```
bterlson, line 40: EnumDeclaration

packages/python/src/components/Parameters.tsx
```diff
@@ -0,0 +1,64 @@
...
+export interface NamedParameter {
+  name: string;
+  type?: Children;
+  defaultValue?: Children;
+}
+
+/**
+ * Render a single parameter as a Declaration (for symbol creation) or as *args/**kwargs.
+ */
+function Parameter(param: NamedParameter) {
```
bterlson, line 14: I think this should have a more specific name since there is a notion of type parameters.

packages/python/src/components/Enum.tsx
mauriciogardini: Fixed.

packages/python/src/components/Value.tsx
mauriciogardini: Fixed.

packages/python/src/components/Method.tsx
mauriciogardini: Fixed.

packages/python/src/components/Class.tsx
mauriciogardini: Fixed.

packages/python/src/components/CallSignature.tsx
```diff
@@ -0,0 +1,243 @@
...
+        <group>
+          <Show when={parameters.length > 0}>, </Show>
+          **kwargs
+        </group>
+      : null}
+    </group>
+  );
+}
+
+function parameter(param: DeclaredParameterDescriptor) {
```
bterlson, line 80: We don't have to export this but I'd make it a JSX component.

packages/python/test/vitest.d.ts
timotheeguerin, line 1: you should be able to remove this one and intead add `@alloy-js/core/testing/matchers` to teh types in the tsconfig.json

timotheeguerin: also adding a vitest.setup.ts so you can actually call `toRenderTo`

swatkatz: done

packages/python/src/components/FunctionDeclaration.tsx
```diff
@@ -0,0 +1,123 @@
...
+ * This will generate:
+ * ```python
+ * def __init__(self: MyClass) -> None:
+ *     self.attribute = "value"
+ * ```
+ *
+ * @remarks
+ *
+ * This is a convenience component that sets the name to `__init__`, marks it as
+ * an instance function, and forces the name to be `__init__` without applying
```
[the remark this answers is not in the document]
mauriciogardini: Created ClassDoc and FunctionDoc with style param. The only valid value for now is "google", but it could be extended to support other styles in the future.

packages/python/src/components/PyDoc.tsx
```diff
@@ -0,0 +1,293 @@
...
+import { Children } from "@alloy-js/core/jsx-runtime";
+import { ParameterDescriptor } from "../parameter-descriptor.js";
+import { Value } from "./index.js";
+
+interface GoogleStyleDocParamTypeProps {
+  type?: Children;
+  optional?: boolean;
+}
+
+function GoogleStyleDocParamType(props: GoogleStyleDocParamTypeProps) {
```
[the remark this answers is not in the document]
mauriciogardini: Created ClassDoc and FunctionDoc with style param. The only valid value for now is "google", but it could be extended to support other styles in the future.

packages/python/src/components/PyDoc.tsx
```diff
@@ -0,0 +1,293 @@
...
+
+/**
+ * A PyDoc comment. The children of this component are joined with two hard
+ * linebreaks. This is useful for creating PyDoc comments with multiple paragraphs.
+ */
+export function PyDoc(props: PyDocProps) {
+  return (
+    <>
+      {'"""'}
+      <align string="">
```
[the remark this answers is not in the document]
mauriciogardini: Indeed, unneeded, removed it.

PR #157, opened 2025-05-14
Various symbol-related refactors.

bterlson:
First, I would just like to apologize. The scope of this PR is larger than I intended, and it has taken longer than I'd have hoped. I fell into this hole some days back after needing to implement a small thing on symbols. And I found that things were kind of a mess. So, this PR cleans up a lot of symbol related things.

There should be no breaking changes to consumers of any of the alloy language libraries, which anyone doing work directly on symbols is likely impacted.

The larger updates are as follows:

## Symbol take and emit

Generally data flows down the render tree. But symbols are a case where we want to flow back "up". Components can now declare that they take a symbol, and handle when a symbol is taken (e.g. by copying or moving those symbol's members). Components can also emit a symbol, which will look up for a component which takes a symbol and give the symbol to that component.

This is a generalization of the AssignmentTarget context. Before you would get the AssignmentTarget context and do stuff on that symbol. Now you just emit the symbol and the parent component picks it up. This allows for references to emit the symbol they resolve to, which in turn allows for things like `<VarDeclaration name="foo">{refkeyForSomeOtherValue}</VarDeclaration>` to track symbols properly,

## OutputScope and OutputSymbol classes

The previous classless approach just wasn't scaling. Different languages implementing their own "subtypes" by passing opaque bags of properties through to the binder was brittle and annoying. And these types were growing various methods and other bits of complexity that was crying out for some encapsulation. So, `binder.createSymbol` is replaced with `new OutputSymbol` or one of the language specific subtypes. Same for scopes.

These new classes have a number of nice benefits:

* The classes are reactive without the need for the reactive proxy, which significantly reduces overhead and needless signal tracking/inadvertant signal firing.
* Setters enforce invariants, e.g. that symbols with a member container flag have the corresponding scope.
* Many properties that couldn't be changed without bad things happening before are now read-only.

## SymbolTable

This new type is a subclass of Set which allows for adding reactive subsets. When a subset is added, any changes to the subset are reflected in the parent, and can be configured to run custom code when this happens. This allows for symbols from one scope to be moved or copied into another scope quite trivially when doing things like instantiation.

## Tracer

You can now set `ALLOY_TRACE` to values like `symbol,scope.create,resolve` to trace happenings in those areas. When a trace event appears, it will include a debugger ID (DID). You can then set `ALLOY_BREAK_ON_DID=12,54,123` to break when those things happen. Want to know why a symbol is being changed? Trace `symbol.update`, find the updates you're curious about, set `ALLOY_BREAK_ON_DID`, and debug.


packages/core/src/symbols/symbol-flow.ts
```diff
@@ -0,0 +1,94 @@
...
+    });
+  } else {
+    ownerContext.takenSymbols!.add(symbol);
+    onCleanup(() => {
+      context!.takenSymbols!.delete(symbol);
+    });
+  }
+}
+
+export function moveTakenMembersTo(baseSymbol: OutputSymbol) {
```
joheredi, line 76: The name of this function doesn't make it clear that it "activates" `takeSymbols`, I was initially confused about why the var declaration component was moving taken symbols but was not signing up for taking symbols first

packages/core/src/symbols/symbol-flow.ts
```diff
@@ -0,0 +1,94 @@
+import {
+  Context,
+  effect,
+  getContext,
+  onCleanup,
+} from "@alloy-js/core/jsx-runtime";
+import { isRef, Ref, shallowReactive } from "@vue/reactivity";
+import { MemberScopeContext, ScopeContext } from "../index.browser.js";
```
joheredi, line 8: is importing from .browser.js intentional?

packages/core/src/symbols/symbol-flow.ts
```diff
@@ -0,0 +1,94 @@
...
+  return context.takenSymbols;
+}
+
+export function emitSymbol(
+  symbol: OutputSymbol | Ref<OutputSymbol | undefined>,
+) {
+  let ownerContext: Context | undefined;
+  let context = getContext()!.owner;
+  while (context) {
+    if (context.takesSymbols) {
```
joheredi, line 32: This is a bit confusing to read. Assigning context to getContext().owner but having `ownerContext` above

packages/core/src/symbols/symbol-flow.ts
```diff
@@ -0,0 +1,94 @@
...
+      ownerContext = context;
+      break;
+    }
+
+    if (
+      context.context &&
+      (context.context[ScopeContext.id] ||
+        context.context[MemberScopeContext.id])
+    ) {
+      // don't cross scope boundaries.
```
joheredi, line 42: If this is something we don't think will be hit normally, maybe a trace/warn would help pinpoint "unexpected" behavior related to this?

packages/typescript/src/components/ObjectExpression.tsx
```diff
@@ -110,11 +108,15 @@ export function ObjectProperty(props: ObjectPropertyProps) {
...
-      name: props.name,
-      refkey: props.refkey,
+    sym = new TSOutputSymbol(props.name, {
+      refkeys: props.refkey,
       flags: OutputSymbolFlags.StaticMember,
     });
+
+    moveTakenMembersTo(sym);
+  } else {
+    // noop
```
joheredi, line 118: Why is takeSymbols() here a noop?

joheredi (approved):

packages/typescript/src/components/ObjectExpression.tsx
bterlson: Since we are presently conditionally making the symbol when we have a refkey and a name, we don't need to do any work when we take a symbol, but we still need to take them otherwise the symbols might end up in a parent somewhere would be incorrect.

packages/core/src/symbols/symbol-flow.ts
bterlson: No sir, stupid vscode.

packages/core/src/symbols/symbol-flow.ts
bterlson: No this will happen all the time when a symbol is emitted but nothing is around to take it. A trace seems appropriate tho.

packages/core/src/symbols/symbol-flow.ts
bterlson: Renamed to symbolTaker.

timotheeguerin (approved):

PR #159, opened 2025-05-15
Upgrade api extractor

timotheeguerin:
Tentative pr. Api extractor just doesn't handle self imports at all well with this new version so have to make sure `@alloy-js/jsx-runtime` is never included


bterlson:
Will do as part of symbol refactor PR.

PR #177, opened 2025-06-12
Add debug tools

bterlson:
This PR introduces debug tools. With these tools you can:

* Inspect symbols and scopes
* Inspect rendered file content, including mapping content to the component which generated it.
* Inspect components
* Rerender components (e.g. to trigger a breakpoint in that component)
* Debug errors with a full component stack
* And more!

As part of this PR, some additional capabilities were added and bugs fixed.


packages/core/src/reactivity.ts
```diff
@@ -112,18 +124,42 @@ export function untrack<T>(fn: () => T): T {
   return v;
 }
 
-export function memo<T>(fn: () => T, equal?: boolean): () => T {
+export interface MemoOptions<T> extends ReactiveEffectOptions<T> {
```
jongio, line 127: The signature change from `memo(fn, equal?: boolean)` to `memo(fn, options?)` is a breaking change. Callers passing `true` as the second arg will now get `options = true` which won't match `options.equal`. Consider an overload or deprecation period.

packages/core/src/scheduler.ts
```diff
@@ -1,45 +1,62 @@
...
 }
 
 export function flushJobs() {
-  let job;
+  if (flushing) {
+    return;
+  }
+  trace(
+    TracePhase.scheduler.flush,
+    () => `flushing jobs, queue size: ${queue.size}`,
```
jongio, line 40: The `flushing` guard prevents re-entrant flushes, which is good. But if an effect queues a new job during flush, that job won't run until the next explicit `flushJobs()` call. Previously the loop would naturally pick it up since `takeJob()` kept pulling from both queues. Worth confirming this doesn't cause stale renders in practice.

jongio (commented):
This branch is significantly behind main and has merge conflicts. Notably, main already includes a more mature devtools implementation (PRs #356 and #398) with configurable ports, non-blocking connection handling, and performance gating via `isDevtoolsEnabled()`. If this PR is still relevant, it likely needs a full rebase and reconciliation with that existing work.

Core concerns if this continues:

**Breaking API changes to `memo()` and `effect()`**: Both functions changed their second parameter from a positional value (`equal?: boolean` / `current?: T`) to an options object. Any downstream consumer passing a boolean or value directly will break silently (the argument becomes the options object, which is truthy). These need either a migration path or a major version bump.

**`flushJobs()` removed from `render()` entry point** (render.ts ~line 188): Previously, effects queued during initial render were flushed before the print phase ran. Removing this changes execution semantics. Reactive updates that depend on synchronous flushing after render will now remain pending, potentially causing stale output. This is likely intentional (the scheduler refactoring moves toward explicit sync/post flush modes) but warrants a note in the PR description.

**Debug server blocks module evaluation**: The top-level `await` in `debug.ts` halts all downstream code until a browser client connects to port 8080. Any process that imports `@alloy-js/core` with `ALLOY_DEBUG` set will hang indefinitely if no client connects. The main branch implementation (`devtools-server.ts`) handles this more gracefully with timeouts and lazy initialization.

**Memory growth in debug mode**: `nodeIds` (Map<number, node>) holds strong references to every render tree node ever created, preventing GC. `fileContents` also grows without bounds. For long-running debug sessions this will eventually exhaust memory.

**Stray `console.log(filePath)`** in the HTTP static file handler (debug.ts). Should be removed or gated behind a verbose flag.

**Commit hygiene**: All 4 commits are "wip" or "rename". If this continues, an interactive rebase into atomic, descriptive commits would help reviewability.

PR #182, opened 2025-06-13
CSharp class method improvemnts

### LINKED ISSUE #179: CS method async support

### LINKED ISSUE #181: CS class method not named with pascal case

### LINKED ISSUE #183: CS abstract class method include body `{}`

timotheeguerin:
fix #179 (Async support
fix #181 (Not respecting name policy)
fix #183 (Abstract method include a body)

### DIFF
--- a/packages/csharp/src/components/ClassMethod.tsx
+++ b/packages/csharp/src/components/ClassMethod.tsx
@@ -31,7 +31,6 @@ export interface ClassMethodProps {
 
   /**
    * If true, the method will be declared as an async method.
-   * If the method has a return type, it will be wrapped in a `Task` automatically.
    */
   async?: boolean;
 }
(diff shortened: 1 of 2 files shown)


packages/csharp/src/components/ClassMethod.tsx
```diff
@@ -0,0 +1,77 @@
...
+import { CSharpOutputSymbol } from "../symbols/csharp-output-symbol.js";
+import { CSharpMemberScope, useCSharpScope } from "../symbols/scopes.js";
+import { ParameterProps, Parameters } from "./Parameters.jsx";
+
+// properties for creating a method
+export interface ClassMethodProps {
+  name: string;
+  refkey?: Refkey;
+  children?: Children;
+  accessModifier?: AccessModifier;
```
joheredi, line 27: What do you think about following a bool prop pattern for accessModifiers similar to the TypeScript compoentns? I like how ergonomic it is to just say <ClassMethod public static ... /> rather than <ClassMethod accessModifiers=["public"] />

Not in the scope of this PR but something to think about.

joheredi: I think we can have better docs this way, like we can document the `async` bool prop and say if you are declaring an async `void` there is no need to pass the return prop...

joheredi: oh wait, async is already a bool prop

joheredi (approved):

packages/csharp/src/components/ClassMethod.tsx
bterlson: +1 bool props is the way.

PR #198, opened 2025-06-17
Add cs interfaces

### LINKED ISSUE #173: [Core] Allow Block to configure body linebraks
There is no way to configure the block component to generate this:

```csharp
public string Foo {get; set;}
```

It will force formatting so the body of the block is in a new line

```csharp
public string Foo {
  get; set;
}
```

### LINKED ISSUE #196: Interface support

timotheeguerin:
fix #173 (inline block)
fix #196 (interface support)

### DIFF
--- a/packages/csharp/src/components/interface/declaration.test.tsx
+++ b/packages/csharp/src/components/interface/declaration.test.tsx
@@ -1,0 +1,45 @@
+import { describe, expect, it } from "vitest";
+import { TestNamespace } from "../../../test/utils.jsx";
+import { InterfaceDeclaration } from "./declaration.jsx";
+
+it("declares class with no members", () => {
+  expect(
+    <TestNamespace>
+      <InterfaceDeclaration name="TestInterface" />
+    </TestNamespace>,
+  ).toRenderTo(`
+      interface TestInterface;
+  `);
+});
+
+describe("modifiers", () => {
+  it.each(["public", "private", "internal"])("%s", (mod) => {
+    expect(
+      <TestNamespace>
+        <InterfaceDeclaration {...{ [mod]: true }} name="TestInterface" />
+      </TestNamespace>,
+    ).toRenderTo(`
+        ${mod} interface TestInterface;
+    `);
+  });
+
+  it.each(["partial"])("%s", (mod) => {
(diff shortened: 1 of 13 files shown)


packages/csharp/src/components/interface/declaration.tsx
timotheeguerin, line 1: what do we think of the naming convention for the files here

packages/csharp/src/components/interface/method.tsx
```diff
@@ -0,0 +1,79 @@
...
+  const params =
+    props.parameters ? <Parameters parameters={props.parameters} /> : "";
+
+  const modifiers = computeModifiersPrefix([
+    getAccessModifier(props),
+    getMethodModifier(props),
+  ]);
+  // note that scope wraps the method decl so that the params get the correct scope
+  return (
+    <MemberDeclaration symbol={methodSymbol}>
```
timotheeguerin, line 69: should we not have a csharp `MemeberDeclaration` type that just take the name instead?

packages/csharp/src/components/interface/declaration.test.tsx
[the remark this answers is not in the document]
bterlson: Feels like there will be SO MANY FILES, but.... I like it, if we can split up into subfolders like this. Let's do it!

packages/csharp/src/components/interface/method.test.tsx
```diff
@@ -0,0 +1,104 @@
...
+  <TestNamespace>
+    <InterfaceDeclaration public name="TestInterface">
+      {props.children}
+    </InterfaceDeclaration>
+  </TestNamespace>
+);
+
+describe("modifiers", () => {
+  describe("access modifiers", () => {
+    it.each(["public", "private", "protected", "internal"] as const)(
```
bterlson, line 17: TIL it.each!

packages/csharp/src/components/interface/declaration.test.tsx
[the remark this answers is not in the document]
timotheeguerin: yeah i like that specially for UI style structure very common to need a lot of files so easier than matching the structure again

PR #209, opened 2025-06-20
Add support for class properties

timotheeguerin:


packages/csharp/src/components/class/property.tsx
```diff
@@ -0,0 +1,127 @@
...
+   * @example `<ClassProperty name="My" get set init={42} />`
+   *
+   * ```cs
+   * int My { get; set; } = 42;
+   * ```
+   */
+  init?: Children;
+}
+
+// a C# interface property
```
bterlson, line 85: Make this a JSDoc comment for doc rendering purposes?

bterlson (approved):

PR #229, opened 2025-06-29
Local references

timotheeguerin:


packages/core/src/symbols/local.test.tsx
```diff
@@ -0,0 +1,128 @@
...
+  function Sub(props: { refkey: Refkey }) {
+    return code`
+      ${props.refkey} ++;
+    `;
+  }
+  expect(
+    <TestWrapper>
+      {code`const ${local("foo")} = 1;`}
+      <hbr />
+      <Sub refkey={local("foo") as any} />
```
timotheeguerin, line 75: this needs to be solved, this should be a refkey assignable

bterlson:
`decl` is in TypeScript, though needs to be added to other languages.

PR #241, opened 2025-07-17
core: Stringify event keys when logging event traces

witemple-msft:
This prevents a crash if the key is a symbol, which cannot be interpolated into a string template (but puzzlingly can be converted to a string using the `String` function).


packages/core/src/reactivity.ts
```diff
@@ -159,12 +159,12 @@ export function effect<T>(fn: (prev?: T) => T, current?: T) {
...
       onTrack(event) {
         trace(TracePhase.effect.track, () => {
-          return `tracking ${event.target}, ${event.key}`;
+          return `tracking ${event.target}, ${String(event.key)}`;
         });
       },
       onTrigger(event) {
         trace(TracePhase.effect.trigger, () => {
-          return `triggering ${event.target}, ${event.key}`;
+          return `triggering ${event.target}, ${String(event.key)}`;
```
timotheeguerin, line 167: This is not the same?

timotheeguerin: TIL

PR #245, opened 2025-07-22
core: Custom element objects

witemple-msft:
This PR adds a new type of child, a CustomChildElement, that is defined by the presence of a well-known symbol property.

During rendering, this type of child is normalized into a function by binding the symbol prop to the child element itself.


packages/core/test/rendering/basic.test.tsx
timotheeguerin, line 1: whats the use case for this?

witemple-msft: The use case is to be able to create objects that act as children in the tree. These can be class instances that carry additional metadata, for example, but have some essence of component-ness within them.

I have a thing I've been using called an "ExpressionBuilder" that allows me to compose syntax using a fluent-style API that is highly type-driven. Think like, `StringShape.literal("asdf").encodeBuffer("utf-8")` composes an expression that takes the literal string "asdf" and encodes it into a Buffer object, but it doesn't actually _do_ that computation, it builds syntax that will do that in the output tree. But when I want to insert the built expression into the alloy tree, I have to destructure it first in a kind of complicated way, because it has a "preamble" part and then an expression part. What I want to do is be able to make the ExpressionBuilder itself usable as if it were an Alloy component, so for example if I want to call an expression I've built up, I can just say `<ts.FunctionCall target={expressionBuilderInstance} />` without having to wrap it in a special "destructure ExpressionBuilder" component. This thing is really designed such that it should be usable as an expression and having a little protocol like this in Alloy to be able to just implement the behavior of rendering an ExpressionBuilder directly is a little bit more convenient than having to wrap everywhere I want to insert one of these things into the tree.

timotheeguerin: Is that in a way a more generic version of this https://github.com/alloy-framework/alloy/pull/237

PR #253, opened 2025-07-25
Add templating components

bterlson:
This PR introduces a few new components:

* CopyFile - copies an existing file into the output directory
* AppendFile - Appends to a file, or specific regions within a file, rather than replacing it
* TemplateFile - Writes a file by interpolating template variables in a template file
* UpdateFile - Updates a file's on-disk contents instead of replacing it

This required a lot of plumbing work and adding file io and such. See changelog entries for details.


pnpm-lock.yaml
```diff
@@ -405,7 +405,123 @@ importers:
         version: 5.8.3
       vitest:
         specifier: 'catalog:'
-        version: 3.2.4(@types/debug@4.1.12)(@types/node@20.17.30)(tsx@4.19.3)(yaml@2.7.1)
+        version: 3.2.4(@types/debug@4.1.12)(@types/node@22.16.5)(jiti@2.5.0)(lightningcss@1.30.1)(tsx@4.20.3)(yaml@2.7.1)
+
+  packages/debug-ui-vibe:
```
timotheeguerin, line 410: was that supposed to be reverted?

packages/core/src/host/interface.ts
```diff
@@ -0,0 +1,161 @@
+/**
+ * Interface for the Alloy host file system operations.
+ *
+ * This interface abstracts file system operations to allow different
+ * implementations across different environments (Node.js, browser, etc.).
+ */
+export interface AlloyHostInterface {
```
timotheeguerin, line 7:
```suggestion
export interface AlloyHost
```

does it need the `Interface` suffix there?

pnpm-lock.yaml
bterlson: blah I dunno, I just did pnpm install is all. I'll rebase main and see what I get.

timotheeguerin (approved):

PR #272, opened 2025-08-31
Golang support for alloy

dimmyjing:
This requires the symbol refactor to be merged first, so it's a draft PR for now.


LarsArtmann:
Hi @dimmyjing, I was looking for this.
Do you have any timeline when it will be ready for review, already?

packages/go/src/symbols/go.ts
```diff
@@ -0,0 +1,228 @@
...
+}
+
+export type GoSymbolKinds =
+  | "symbol"
+  | "named-type"
+  | "function"
+  | "field"
+  | "package";
+
+function ensureNameExport(
```
bterlson, line 30: I feel like we talked about this on discord but I'm not the biggest fan of tracking pointerness in the symbol name like this. Another approach might be that the symbol itself can store a flag for if its holding a pointer, and your member expression component equivalent can dereference when needed based on this metadata. I imagine there are cases when member expression will need to do the opposite and & a value-bearing symbol so you might need that machinery anyway?

bterlson: It would also simplify this logic a lot, and decouple name from pointerness, simplifying the below code.

packages/go/src/symbols/go.ts
```diff
@@ -0,0 +1,228 @@
...
+        );
+        if (oldName !== newName) {
+          this.name = newName;
+        }
+      },
+    );
+  }
+
+  #pointer: boolean = false;
+  get pointer() {
```
bterlson, line 129: Symbols have been generally using `is` prefix for boolean props.

packages/go/src/symbols/go.ts
```diff
@@ -0,0 +1,228 @@
...
+  track,
+  TrackOpTypes,
+  trigger,
+  TriggerOpTypes,
+  watch,
+} from "@alloy-js/core";
+import { GoScope } from "../scopes/go.js";
+import { PackageSymbol } from "./package.js";
+
+export interface GoSymbolOptions extends OutputSymbolOptions {
```
bterlson, line 17: Doc comments here appreciated.

packages/go/src/symbols/go.ts
```diff
@@ -0,0 +1,228 @@
...
+    const old = this.#pointer;
+    if (old === value) {
+      return;
+    }
+    this.#pointer = value;
+    trigger(this, TriggerOpTypes.SET, "pointer", value, old);
+  }
+
+  #canExport: boolean;
+  get canExport() {
```
bterlson, line 143: Doc comments on each of these accessors.

packages/go/src/symbols/go.ts
```diff
@@ -0,0 +1,228 @@
...
+    if (this.spaces.length === 0) {
+      return undefined;
+    }
+
+    // todo: probably need to validate that a symbol can't belong to spaces in
+    // multiple packages.
+    const firstSpace = this.spaces[0];
+
+    if (firstSpace instanceof OutputMemberSpace) {
+      // this symbol is a member of something, so get the enclosing namespace from
```
bterlson, line 180:
```suggestion
      // this symbol is a member of something, so get the enclosing package from
```

packages/go/src/symbols/go.ts
```diff
@@ -0,0 +1,228 @@
...
+      ...this.getCopyOptions(),
+      pointer: this.#pointer,
+      canExport: this.#canExport,
+      exported: this.#exported,
+    };
+  }
+
+  protected initializeGoCopy(copy: GoSymbol) {
+    this.initializeCopy(copy);
+    watch(
```
bterlson, line 208: I think this is missing canExport?

packages/go/src/symbols/named-type.ts
```diff
@@ -0,0 +1,125 @@
...
+  | "struct-member"
+  | "interface"
+  | "interface-member"
+  | "package";
+export type NamedTypeSymbolKind = "named-type" | "package";
+
+/**
+ * Information for a Go generic type parameter.
+ */
+export interface TypeParameterProps {
```
bterlson, line 29: Move this with the component?

packages/go/src/components/pointer/pointer.tsx
```diff
@@ -0,0 +1,52 @@
...
+  const symbols = ref<Set<OutputSymbol>>(new Set()) as Ref<
+    Set<OutputSymbol>,
+    Set<OutputSymbol>
+  >;
+
+  effect(() => {
+    if (symbols.value.size !== 1) return;
+    const symbol = Array.from(symbols.value)[0];
+    if (!(symbol instanceof NamedTypeSymbol)) return;
+    const pointerSymbol = symbol.pointerSymbol;
```
bterlson, line 33: It seems that this implies that all pointers to a particular type share a symbol. This doesn't seem accurate. For example, there could be a name conflict in some scope so the name of the base type is different? I would make a transient symbol whose type is set to the pointer type.

packages/go/src/components/pointer/pointer.tsx
```diff
@@ -0,0 +1,52 @@
...
+  });
+
+  return (
+    <PointerContext.Provider>
+      <PointerInner symbols={symbols}>*{props.children}</PointerInner>
+    </PointerContext.Provider>
+  );
+}
+
+function PointerInner(props: {
```
bterlson, line 45: What is this for? Feels like the `Pointer` component can take the symbols.

bterlson: And if not, you could use `createSymbolSlot`. Its new `firstSymbol` property is probably what you'd want to use.

packages/go/src/symbols/named-type.ts
```diff
@@ -0,0 +1,125 @@
...
+      this.name,
+      undefined,
+      this.#typeKind,
+      options,
+    );
+    this.initializeGoCopy(copy);
+    return copy;
+  }
+
+  #pointerSymbol?: GoSymbol;
```
bterlson, line 108: Noted above but it seems problematic to have a single pointer symbol for a given type. Each type can have multiple pointers in different scopes no?

packages/go/src/components/ImportStatement.tsx
```diff
@@ -0,0 +1,115 @@
...
+  });
+
+  const rendered = memo(() => {
+    if (imports.value.length > 1) {
+      return (
+        <>
+          import (
+          <Indent hardline>
+            <For each={imports}>
+              {(imp) => (imp === null ? <></> : <ImportStatement {...imp} />)}
```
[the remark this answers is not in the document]
dimmyjing: nulls are intentional empty lines, in go the convention is to import builtins, newline, third party dependencies, newline, and then local dependencies, so I have nulls in imports to signal newlines in this case

packages/go/src/symbols/go.ts
dimmyjing: the exportness and pointerness are stored separately, although I couldn't see any good ways of automatically changing the displayed name based on the pointerness so I'm using that function to ensure the name follows the exportness and pointerness whenever anything changes. If there's another way to display a separate name from the original one then that would be good too.

packages/go/src/components/pointer/pointer.tsx
dimmyjing: my goal for this is to essentially treat the pointer type as a copy of the actual type. the reason is because it is possible to define methods on a pointer type but not the original type, so the only way to call that method is to call it on the pointer type. I thought it would be easier to assign the method onto a concrete pointer type so other instances that is of the pointer type can reference those methods.

packages/go/src/symbols/named-type.ts
dimmyjing: If I leave TypeParameterProps in `typeparameters.tsx` then `pnpm build` complains that `The symbol "TypeParameterProps_2" needs to be exported by the entry point index.d.ts` in `named-type.ts`. I'm not sure why that would error since `TypeParameterProps` is exported from `index.ts`

packages/go/src/components/pointer/pointer.tsx
dimmyjing: it is implemented using `createSymbolSlot` and `firstSymbol` now but you should check if it's used correctly.

samples/go-example/src/index.tsx
```diff
@@ -0,0 +1,184 @@
...
+  code,
+  List,
+  Output,
+  refkey,
+  render,
+  writeOutput,
+} from "@alloy-js/core";
+import * as go from "@alloy-js/go";
+
+const aRef = refkey("a");
```
bterlson, line 11: Update these to `namekey` and you can avoid passing name to all the props!

packages/go/src/components/struct/declaration.tsx
```diff
@@ -0,0 +1,230 @@
...
+   * This is not common in Go, so use with caution.
+   */
+  singleLine?: boolean;
+}
+
+/**
+ * Wrapper for creating a named struct type declaration.
+ */
+export function StructTypeDeclaration(
+  props: StructDeclarationProps & TypeDeclarationProps,
```
bterlson, line 48: Let's use interface extends here (it will work better with the documentation system).

packages/go/src/components/parameters/typeparameters.tsx
```diff
@@ -0,0 +1,51 @@
+import { For, Indent, MemberDeclaration, MemberName } from "@alloy-js/core";
```
bterlson, line 1: type-parameters.tsx

PR #282, opened 2025-09-04
More detailed unresolved symbol

### LINKED ISSUE #195: <Unresolved Symbol> can it show the name of the symbol
Would make it easier to debug what didn't resolve

timotheeguerin:
fix  #195 
Show the refkey information for easier debugging


packages/typescript/src/symbols/reference.tsx
```diff
@@ -32,7 +34,7 @@ export function ref(
 
   return memo(() => {
     if (resolveResult.value === undefined) {
-      return ["<Unresolved Symbol>", undefined];
+      return [`<Unresolved Symbol: ${getRefkeyString(refkey)}>`, undefined];
```
timotheeguerin, line 37: as every language needs this, should core provide an helper for that piece of code or do we see some language wanting a specific syntax so maybe it doesn't fail compilation(if thats ever wanted)

PR #285, opened 2025-09-05
Remove extra linebreak from core Block and python PythonBlock components

bterlson:


packages/core/src/components/Block.tsx
```diff
@@ -10,9 +10,10 @@ export interface BlockProps {
   opener?: string;
 
   /**
-   * The closing punctuation of the block. Defaults to "\}".
+   * The closing punctuation of the block. Defaults to "\}". When passed `false`,
+   * no closing punctuation will be added and there will be no trailing newline.
    */
-  closer?: string;
+  closer?: string | boolean;
```
jongio, line 16: Consider narrowing this to string | false instead of string | boolean. Only alse has defined special semantics here (suppress trailing break + render nothing). Passing 	rue would silently produce no visible closer but still emit a trailing break, which is an odd state that's hard to distinguish from the default. string | false communicates intent more precisely and prevents accidental misuse.

.chronus/changes/block-closer-2025-8-4-19-4-17.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: feature
+packages:
+  - "@alloy-js/core"
+---
+
+The `<Block>` component no longer emits an empty line when the `closer` prop is `false.
```
jongio, line 7: Minor: missing closing backtick after alse - should be ` alse ` with the closing tick.

jongio (commented):
The design makes sense: distinguishing "no visible closer" (closer="") from "no closer AND no trailing break" (closer={false}) is a clean way to fix the extra blank lines Python blocks were emitting. Tests are thorough and the migration path is documented in the chronus changefiles.

Two small items below.

Note: this PR has merge conflicts with main that'll need resolving before merge.

PR #302, opened 2025-09-16
Execute changes in the Python implementation

mauriciogardini:
## Create `SingleTypeExpression`
Create `TypeReference ` to handle some edge cases on typing (involving typeArgs) to, in addition to the already present `UnionTypeExpression`, more properly represent the typing in Python.

## Refactor `FunctionDeclaration`
Refactor FunctionDeclaration to more properly represent each scenario:

- FunctionDeclaration: Standalone function. Has the functionType parameter.
- MethodDeclaration: Normal method
- ClassMethodDeclaration: Class method, prepended with @classmethod
- StaticMethodDeclaration: Static method, prepended with @staticmethod
- DunderMethodDeclaration: Any dunder method. Any formatting to the name that would normally happen will be skipped
- ConstructorDeclaration: Special component for the __new__ dunder method, which has cls as its first argument, but isn't decorated with @classmethod
- PropertyDeclaration: Python properties, with setter and deleter support, prepended with @property

Also:
- Added abc module to be able to import @staticmethod
- Modified createPythonSymbol so it supports a reuseExisting, which looks for existing symbol instead of creating a new one. Intended to handle the @property special cases

## Split Enum implementations
Split Enum implementations into FunctionalEnumDeclaration and ClassEnumDeclaration.

## Split docs components into separate components
Make docs components more granular by splitting them into:
- ClassDoc
- FunctionDoc
- MethodDoc
- ModuleDoc
- PropertyDoc
- GeneratorDoc
- ExceptionDoc
- AttributeDoc

These are the only components that are meant to be used directly, and are the only ones that are exported.

Also added the equivalent GoogleStyle-prefixed classes, bringing the Google-style implementation of these.


packages/python/src/components/CallSignature.tsx
```diff
@@ -1,24 +1,20 @@
...
   Show,
   SymbolSlot,
-  useContext,
 } from "@alloy-js/core";
 import { ParameterDescriptor } from "../parameter-descriptor.js";
 import { createPythonSymbol } from "../symbol-creation.js";
 import { PythonOutputSymbol } from "../symbols/index.js";
 import { Atom } from "./Atom.jsx";
-import { PythonSourceFileContext } from "./SourceFile.jsx";
+import { type TypeExpressionProps } from "./index.js";
```
bterlson, line 12: Best not to import from index. It will work but there will come a time when you are bit by circular reference problems and it will be awful. Ask me how I know 😂

packages/python/src/components/CallSignature.tsx
```diff
@@ -202,7 +156,7 @@ export interface CallSignatureProps {
  * ```tsx
  * <CallSignature
  *   parameters={[{ name: "a", type: "int" }, { name: "b", type: "str" }]}
- *   returnType="int"
+ *   returnType={"int"}
```
bterlson, line 159: hopefully these are equivalent.

packages/python/src/components/CallSignature.tsx
```diff
@@ -179,20 +143,10 @@ export interface CallSignatureProps {
...
-  /**
-   * Indicates that this is a class function.
-   */
-  classFunction?: boolean; // true if this is a class function
-
   /**
    * The return type of the function.
    */
-  returnType?: Children;
+  returnType?: TypeExpressionProps;
```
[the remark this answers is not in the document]
mauriciogardini: Will remove TypeExpressionProps across the repo and replace for Children.

packages/python/src/components/FunctionDeclaration.tsx
```diff
@@ -1,101 +1,653 @@
...
+        instance:
+          props.functionType !== undefined && currentScope?.isMemberScope,
+        refkeys: props.refkey,
+      },
+      "function",
+    );
+
+  // Only emit symbol if we created it (not if using existing one)
+  if (!props.sym) {
+    emitSymbol(sym);
```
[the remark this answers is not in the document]
mauriciogardini: So, this is a very specific scenario regarding properties, implemented in `PropertyDeclaration`. Maybe you remember that from our previous discussions.

```
class C:
    def __init__(self):
        self._x = None

    @property
    def x(self):
        """I'm the 'x' property."""
        return self._x

    @x.setter
    def x(self, value):
        self._x = value

    @x.deleter
    def x(self):
        del self._x
```

On this case, we can have up to 3 methods sharing the same name. From our previous discussions, we agreed on giving them all the same symbol, and that's what we are doing with that.

So what's happening on practice is:
1) User creates a property using a `PropertyDeclaration`. A simple example (with no body, just to illustrate) would be:
```
<py.PropertyDeclaration name="x">
  <py.PropertyDeclaration.Setter />
  <py.PropertyDeclaration.Deleter />
</py.PropertyDeclaration>
```

2) PropertyDeclaration will calculate the appropriate information and then render the (up to) 3 methods. This is the "skeleton" of the `PropertyDeclaration` rendering (omitting some parts for the sake of brevity):
```
      <DeclarationContext.Provider value={sym}>
        <PropertyContext.Provider value={props.type}>
          <List hardline enderPunctuation>
            <PropertyMethodDeclaration
              abstract={props.abstract}
              doc={props.doc}
            />
            <PropertyDeclaration.Setter />
            <PropertyDeclaration.Deleter />
          </List>
        </PropertyContext.Provider>
      </DeclarationContext.Provider>
```

3) Going from the top, we will first create a `PropertyMethodDeclaration`, which is going to create the `property` decorated method. It will render a `MethodDeclarationBase`, which will render a `BaseFunctionDeclaration` which will [create a symbol](https://github.com/alloy-framework/alloy/pull/302/files/e55c741e9d51243e6c9bc1f588668fbf4537062b#diff-734db50189b633a46d9757e1c2443d992161776df21d0e0cbdebce4491f3ade0R97). As you already mentioned, `createPythonSymbol` should handle symbol emission just fine
4) The `PropertyMethodDeclaration` is created inside a `DeclarationContext`, something that was also previously suggested, which allows us to share the symbol created by it with other components inside the same context.
5) Then, we will create the setter. Both the `Setter` and the `Deleter` are taggedComponents, and will render a `PropertyMethodBase`. This component takes the symbol from the `DeclarationContext` and [passes it](https://github.com/alloy-framework/alloy/pull/302/files/e55c741e9d51243e6c9bc1f588668fbf4537062b#diff-734db50189b633a46d9757e1c2443d992161776df21d0e0cbdebce4491f3ade0R444) to the creation of a `MethodDeclarationBase`, which will render a `BaseFunctionDeclaration`. The `BaseFunctionDeclaration` would normally create a symbol, but since we are passing one, it just skips symbol creation and just emits it again.
6) The same applies for the deleter.

We are also using a PropertyContext to allow us to share typing information between these components. We could have appended the type information as a metadata in the symbol creation, but I didn't like that idea, and went for the "separate context just to share the type for this very specific scenario related to properties" approach.

packages/python/src/components/FunctionDeclaration.tsx
```diff
@@ -1,101 +1,647 @@
...
-  const sym = createPythonSymbol(
+  return <BaseFunctionDeclaration {...props} />;
+}
+
+/**
+ * Internal method declaration base component that handles validation
+ * and ensures the method is declared within a class (member scope).
+ * This component is not exported to keep implementation details private.
+ */
+function MethodDeclarationBase(
```
bterlson, line 174: Let's try to keep to one component per file as much as is reasonable.

packages/python/src/components/FunctionBase.tsx
```diff
@@ -0,0 +1,88 @@
...
+ *   return x
+ * </BaseFunctionDeclaration>
+ * ```
+ * Generates:
+ * ```python
+ * def helper(x):
+ *     return x
+ * ```
+ */
+export function BaseFunctionDeclaration(props: BaseFunctionDeclarationProps) {
```
bterlson, line 56: Does this even need to be exported?

packages/python/src/components/index.ts
```diff
@@ -2,19 +2,27 @@ export * from "./Atom.js";
...
 export * from "./EnumDeclaration.js";
 export * from "./EnumMember.js";
+export * from "./FunctionBase.js";
 export * from "./FunctionCallExpression.js";
 export * from "./FunctionDeclaration.js";
 export * from "./ImportStatement.js";
 export * from "./LexicalScope.js";
 export * from "./MemberExpression.js";
 export * from "./MemberScope.jsx";
+export * from "./MethodBase.js";
```
bterlson, line 18: This says its internal but it's exported. Intentional?

packages/python/test/callsignatures.test.tsx
```diff
@@ -78,74 +100,45 @@ describe("Call Signature Parameters", () => {
 
 describe("Call Signature", () => {
   it("renders a simple call signature", () => {
-    const result = toSourceText([<py.CallSignature parameters={["a", "b"]} />]);
```
bterlson, line 81: The signature still takes string[], so I hope this change isn't necessary?

PR #324, opened 2025-10-17
TypeSpec SourceFile and Namespace components

jpalvarezl:
Open questions:
- `test/components/namespace.test.tsx` we are targeting to make that unit test pass but nothing renders. We are struggling to find the minimal code necessary for the setup to work (the code has been mostly lifted from the csharp package).

### DIFF
--- a/packages/typespec/src/components/namespace/namespace.tsx
+++ b/packages/typespec/src/components/namespace/namespace.tsx
@@ -24,7 +24,6 @@ export function Namespace(props: NamespaceProps) {
             {props.children}
         </NamespaceContext.Provider>);
     } else {
-        sfScope.hasBlockNamespace = true;
         return (
             <>
                 namespace <NamespaceName symbol={namespaceSymbol} relative />{" "}


packages/typespec/src/components/source-file/source-file.tsx
```diff
@@ -0,0 +1,66 @@
...
+        namespaceSymbol = new NamespaceSymbol(props.namespace);
+    }
+    const namespaceScope = new NamespaceScope(namespaceSymbol, sourceFileScope);
+
+    const options = useTypeSpecFormatOptions();
+
+    return (
+        <CoreSourceFile
+            path={props.path}
+            filetype=".tsp"
```
timotheeguerin: @bterlson did we wanted to start doing mime type for this or just a name like `typespec`?

packages/typespec/test/utils.tsx
```diff
@@ -0,0 +1,76 @@
...
+} from "@alloy-js/core";
+
+
+import * as typespec from "../src/index.js";
+
+import { createTypeSpecNamePolicy } from "../src/name-policy.js";
+import { TypeSpecFormatOptions } from "../src/contexts/format-options.js";
+import { SourceFile } from "../src/index.js";
+
+export function toSourceText(c: Children, options?: PrintTreeOptions): string {
```
timotheeguerin: shouldn't need this in test

packages/typespec/src/components/namespace/file-level.tsx
```diff
@@ -0,0 +1,52 @@
+import { Children, Declaration, Name, Namekey, Refkey, Scope, Show, useBinder } from "@alloy-js/core";
+import { useSourceFileScope } from "../../scopes/source-file.js";
+import { getGlobalNamespace } from "../../contexts/global-namespace.js";
+import { NamespaceSymbol } from "../../symbols/namespace.js";
+import { createNamespaceSymbol } from "../../symbols/factories.js";
+import { NamespaceScope } from "../../scopes/namespace.js";
+
+export interface FileLevelNamespaceProps {
```
bterlson: I think there can just be one Namespace component. What we do in C# is render block namespaces if you have a namespace within a source file, otherwise blockless namespaces when the namespace component is outside a source file. Essentially, if an entire source file is inside some namespace, we always emit the blockless namespace.

packages/typespec/src/components/source-file/source-file.tsx
bterlson: I would like to but I'm not sure, I keep getting hung up on the fact that many languages (including tsp) don't have a registered mime type or even established convention around one. I think this is fine for now.

packages/typespec/test/utils.tsx
```diff
@@ -0,0 +1,76 @@
...
+        {c}
+      </CoreSourceFile>
+    </Output>,
+    { insertFinalNewLine: false, ...options },
+  );
+
+  return findFile(res, "test.tsp").contents;
+}
+
+export function findFile(
```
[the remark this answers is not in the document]
bterlson: Yes, we have evolved the test framework a lot since many of the existing tests were written ☺️

packages/typespec/src/components/source-file/source-file.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { getGlobalNamespace } from "../../contexts/global-namespace.js";
+import { NamespaceScopes } from "#components/namespace-scopes.jsx";
+import { NamespaceName } from "#components/namespace/namespace-name.jsx";
+import { NamespaceSymbol } from "../../symbols/namespace.js";
+
+export interface SourceFileProps {
+    path: string;
+
+    /** If present, it defines a file-level namespace (if not present, it uses the global namespace) */
+    namespace?: NamespaceSymbol;
```
glecaros: probably we could receive a symbol or string here.

packages/typespec/src/scopes/named-type.ts
```diff
@@ -0,0 +1,44 @@
...
+import { NamedTypeSymbol } from "../symbols/named-type.js";
+import { TypeSpecScope } from "./typespec.js";
+import { SourceFileScope } from "./source-file.js";
+
+/**
+ * This scope contains NamedTypeSymbols for types that are declared in
+ * containers like namespaces. This scope is a member scope whose
+ * member symbol is a NamedTypeSymbol.
+ */
+export class TypeSpecNamedTypeScope extends TypeSpecScope {
```
glecaros: do we need this in this PR?

packages/typespec/src/symbols/reference.tsx
```diff
@@ -0,0 +1,81 @@
...
+import { NamespaceScope } from "../scopes/namespace.js";
+import { useSourceFileScope } from "../scopes/source-file.js";
+import { TypeSpecSymbol } from "./typespec.js";
+import { NamespaceSymbol } from "./namespace.js";
+
+// converts a refkey to its fully qualified name
+// e.g. if refkey is for bar in enum type foo, and
+// foo is in the same namespace as the refkey, then
+// the result would be foo.bar.
+export function ref(
```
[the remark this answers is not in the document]
jpalvarezl: It's a cascading dependency necessary to define `Reference.tsx` so we can pass that to `CoreSourceFile` : https://github.com/alloy-framework/alloy/pull/324/files#diff-668821ea9c50226ee835e35b54cddd4afc9313f85249093722a652ad3d763187R51

PR #325, opened 2025-10-21
Add Dataclass component

mauriciogardini:
Add the Dataclass component to the Python implementation. Used the ClassComponent, just applying some specific behavior on top of it.

### DIFF
--- a/packages/python/src/components/DataclassDeclaration.tsx
+++ b/packages/python/src/components/DataclassDeclaration.tsx
@@ -160,11 +160,6 @@ export function DataclassDeclaration(props: DataclassDeclarationProps) {
     validateDataclassDecoratorArgs(kwargs);
   }
 
-  function RunSymbolValidation() {
-    validateDataclassMemberConflicts(kwargs as DataclassDecoratorKwargs);
-    return null;
-  }
-
   return (
     <>
       {"@"}
@@ -183,7 +178,7 @@ export function DataclassDeclaration(props: DataclassDeclarationProps) {
       <hbr />
       <ClassDeclaration name={props.name} bases={props.bases} doc={props.doc}>
         <StatementList>{props.children}</StatementList>
-        <RunSymbolValidation />
+        {validateDataclassMemberConflicts(kwargs as DataclassDecoratorKwargs)}
       </ClassDeclaration>
     </>
   );


packages/python/src/components/ClassDeclaration.tsx
```diff
@@ -53,8 +60,8 @@ export function ClassDeclaration(props: ClassDeclarationProps) {
     "class",
   );
 
-  const hasChildren =
-    childrenArray(() => props.children).filter((c) => Boolean(c)).length > 0;
+  const childrenComputed = computed(() => childrenArray(() => props.children));
+  const hasChildren = childrenComputed.value.some(Boolean);
```
bterlson:
```suggestion
  const hasChildren = computed(() => childrenArray(() => props.children).some(Boolean));
```

packages/python/src/components/UnionTypeExpression.tsx
```diff
@@ -1,18 +1,18 @@
-import { Children, For, childrenArray } from "@alloy-js/core";
+import { Children, For, childrenArray, computed } from "@alloy-js/core";
 
 export interface UnionTypeExpressionProps {
   children: Children;
 }
 
 export function UnionTypeExpression(props: UnionTypeExpressionProps) {
-  const items = childrenArray(() => props.children);
+  const items = computed(() => childrenArray(() => props.children));
```
bterlson:
```suggestion
```

Don't need this

packages/python/src/components/UnionTypeExpression.tsx
```diff
@@ -1,18 +1,18 @@
...
   return (
     <group>
       <ifBreak>(</ifBreak>
       <indent>
         <sbr />
         <For
-          each={items}
+          each={items.value}
```
bterlson: Can use `<List children={props.children} ... rest of args to For ... />`

packages/python/src/components/DataclassDeclaration.tsx
```diff
@@ -0,0 +1,189 @@
...
+    );
+  }
+
+  if (kwargs.order === true && kwargs.eq === false) {
+    throw new Error("order=True requires eq=True in @dataclass decorator");
+  }
+
+  if (kwargs.order === true) {
+    const orderingMethods = ["__lt__", "__le__", "__gt__", "__ge__"];
+    const conflict = findMethodDeclaration(children, orderingMethods);
```
bterlson: I'm not sure this validation is necessary, especially given that this will only find method declarations if they are present as direct children of this component. If I have e.g. a set of methods bundled into some reusable component, this won't find it. I'd say if you want to keep this kind of validation it needs to operate at the symbol level (checking to see if a symbol for this method name already exists).

packages/python/src/components/DataclassDeclaration.tsx
```diff
@@ -0,0 +1,189 @@
...
+  }
+
+  return (
+    <>
+      {"@"}
+      {dataclassesModule["."].dataclass}
+      <Show when={hasDecoratorArgs}>
+        {"("}
+        <For
+          each={Object.keys(kwargs ?? {}).map((k) => [k, (kwargs as any)[k]])}
```
bterlson: I think this is just `Object.entries`?

packages/python/src/components/DataclassDeclaration.tsx
```diff
@@ -0,0 +1,189 @@
...
+ *     id: int
+ *     _: KW_ONLY
+ *     name: str = "Anonymous"
+ * ```
+ */
+export function DataclassDeclaration(props: DataclassDeclarationProps) {
+  const kwargs = props.decoratorKwargs as DataclassDecoratorKwargs | undefined;
+  const hasDecoratorArgs =
+    kwargs !== undefined && Object.keys(kwargs).length > 0;
+  const childrenComputed = computed(() => childrenArray(() => props.children));
```
bterlson: Same mod as above.

packages/python/src/components/DataclassDeclaration.tsx
```diff
@@ -0,0 +1,189 @@
...
+          comma
+          space
+        >
+          {([k, v]) => (
+            <>
+              {k}=<Atom jsValue={v} />
+            </>
+          )}
+        </For>
+        {")"}
```
bterlson:
```suggestion
        )
```

packages/python/src/components/DataclassDeclaration.tsx
```diff
@@ -0,0 +1,189 @@
...
+}
+
+/**
+ * Renders a Python dataclass. Uses ClassDeclaration component internally.
+ *
+ * Example:
+ * ```tsx
+ * <py.DataclassDeclaration name="User" decoratorKwargs={{ kw_only: true }}>
+ *   <py.VariableDeclaration instanceVariable omitNone name="id" type="int" />
+ *   <py.DataclassKWOnly />
```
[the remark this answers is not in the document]
mauriciogardini: Removing DataclassKWOnly and replacing for a normal variable assigment.

packages/python/src/components/PropertyDeclaration.tsx
```diff
@@ -93,7 +94,8 @@ export function PropertyDeclaration(props: PropertyDeclarationProps) {
     }
   };
 
-  const children = childrenArray(() => props.children);
+  const childrenComputed = computed(() => childrenArray(() => props.children));
+  const children = childrenComputed.value;
```
bterlson: This change is a no-op, before is simpler.

packages/python/src/components/DataclassDeclaration.tsx
```diff
@@ -0,0 +1,212 @@
...
+  for (const key of Object.keys(props)) {
+    // Only include known flags; skip undefined values
+    if (validKeySet.has(key)) {
+      const value = (props as any)[key];
+      if (value !== undefined)
+        orderedKwargs.push([key as keyof DataclassDecoratorKwargs, value]);
+    }
+  }
+  // Materialize ordered entries into an object for validation/rendering
+  const kwargs = orderedKwargs.reduce((acc, [k, v]) => {
```
bterlson: Use Object.fromEntries

packages/python/src/components/DataclassDeclaration.tsx
```diff
@@ -0,0 +1,212 @@
...
+    (acc as any)[k] = v;
+    return acc;
+  }, {} as DataclassDecoratorKwargs);
+  const hasDecoratorArgs = orderedKwargs.length > 0;
+  const toSnakeCase = (s: string): string =>
```
bterlson: Use the change-case library for this (it's in the catalog).

packages/python/src/components/DataclassDeclaration.tsx
```diff
@@ -0,0 +1,212 @@
...
+    s
+      .replace(/([a-z\d])([A-Z])/g, "$1_$2")
+      .replace(/([A-Z]+)([A-Z][a-z])/g, "$1_$2")
+      .toLowerCase();
+  const decoratorEntries = orderedKwargs.map(([k, v]) => {
```
bterlson: Recase when you assemble orderedKwargs, it seems like we don't need the camelCased version for anything so might as well do it in one pass.

packages/python/src/components/PyDoc.tsx
```diff
@@ -258,26 +265,25 @@ export interface PyDocExampleProps {
  * Create a PyDoc example, which is prepended by \>\>.
  */
 export function PyDocExample(props: PyDocExampleProps) {
```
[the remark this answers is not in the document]
mauriciogardini: Will add some docstrings to make it more clear; will also make sure we handle non-string children.

packages/python/src/components/DataclassDeclaration.tsx
```diff
@@ -0,0 +1,190 @@
...
+              {k}=<Atom jsValue={v} />
+            </>
+          )}
+        </For>
+        )
+      </Show>
+      <hbr />
+      <ClassDeclaration name={props.name} bases={props.bases} doc={props.doc}>
+        <StatementList>{props.children}</StatementList>
+        <RunSymbolValidation />
```
bterlson: I think you could just do `{validateDataclassMemberConflicts(kwargs as DataclassDecoratorKwargs)}` and avoid the component.

bterlson (approved):

bterlson (approved):

PR #328, opened 2025-10-28
Add various features around empty content handling

bterlson:
The renderer now bookkeeps whether contexts have content underneath them. This capability is used to provide:

* `createContentSlot` which returns a component that exposes whether its children have contents.
* `List`, `For`, etc. now handle list items that don't have any content properly.

I've also updated the python components to use the content slot.


timotheeguerin:
🎉🎉

PR #335, opened 2025-11-06
Add implements support for Typescript classes

FrankShaw:
Noticed that only `extends` was supported for Typescript classes and we wanted to use `implements` for some built-in NestJS interfaces in our project. 

Open to feedback about the implementation but it seemed pretty straightforward

### DIFF
--- a/.chronus/changes/ts-class-implements-2025-10-11-20-47-30.md
+++ b/.chronus/changes/ts-class-implements-2025-10-11-20-47-30.md
@@ -1,0 +1,8 @@
+---
+# Change versionKind to one of: breaking, feature, fix, internal
+changeKind: fix
+packages:
+  - "@alloy-js/typescript"
+---
+
+Add implements support for Typescript classes


packages/typescript/src/components/ClassDeclaration.tsx
```diff
@@ -65,6 +68,17 @@ export interface ClassDeclarationProps extends CommonDeclarationProps {
  */
 export function ClassDeclaration(props: ClassDeclarationProps) {
   const extendsPart = props.extends && <> extends {props.extends}</>;
+  const implementsArray =
+    props.implements && childrenArray(() => props.implements);
```
bterlson: It's best not to use childrenArray if you can avoid it. It makes it hard to compose with other components such as some component which wraps references to a standard set of extends. For other similar things (e.g. args for `FunctionCallExpression` we are using a signature like `Children[]`). It makes the call site a bit less convenient for single-interface extends but it is much less surprising behavior overall.

bterlson (approved):

PR #341, opened 2025-11-27
Feature/go explicit public prop

### LINKED ISSUE #330: [Go] public vs private
Can we support auto-naming logic for public and private symbol?

LarsArtmann:
Fixes #330

Assisted-By: Z.ai GLM-4.6 via Crush

### DIFF
--- a/packages/go/src/symbols/factories.ts
+++ b/packages/go/src/symbols/factories.ts
@@ -223,12 +223,11 @@ export function createAnonymousTypeSymbol(
   );
 }
 
-function withNamePolicy<T extends { namePolicy?: NamePolicyGetter }>(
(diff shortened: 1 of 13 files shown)


packages/go/src/symbols/factories.ts
```diff
@@ -223,12 +223,16 @@ export function createAnonymousTypeSymbol(
   );
 }
 
-function withNamePolicy<T extends { namePolicy?: NamePolicyGetter }>(
-  options: T,
+function withNamePolicy(
+  options: GoSymbolOptions,
```
LarsArtmann: Consider `withNamePolicy<T extends GoSymbolOptions>`

PR #345, opened 2025-12-11
Ability to define peer and dev deps for auto-imported packages

bterlson:


packages/typescript/src/create-package.ts
```diff
@@ -165,10 +165,19 @@ function assignMembers(
   }
 }
 
+const packageScopeSymbol: unique symbol = Symbol();
+/**
+ * Retrieve the package scope associated with an external package created via
+ * createPackage.
+ */
+export function getPackageScope(pgk: ExternalPackage) {
```
jongio, line 173: Typo: `pgk` should be `pkg`.

packages/typescript/src/components/PackageJson.tsx
```diff
@@ -73,8 +74,32 @@ export interface PackageExports {
...
+      peerDependencies: props.peerDependencies,
+    };
+
+    if (pkg) {
+      for (const dependency of pkg.scope.dependencies) {
+        const kind = pkgMeta?.dependencyType.get(dependency) ?? "dependencies";
+        const versionSpecifier =
+          pkgMeta?.versionSpecifiers.get(dependency) ?? dependency.version;
+
+        kinds[kind] ??= {};
```
jongio, line 92: Potential mutation: `kinds[kind]` holds a direct reference to `props.dependencies` (or `devDependencies`/`peerDependencies`). Assigning `kinds[kind][dependency.name] = versionSpecifier` mutates the original props object if it was passed by reference. Consider spreading: `kinds[kind] = { ...(kinds[kind] ?? {}), [dependency.name]: versionSpecifier }` or building a fresh object to avoid side effects inside the `memo`.

jongio (commented):
This PR appears to be superseded by #347 ("Adding package versions + kind configuration support for PackageDirectory") which was merged the same day these commits were pushed. The PR currently has merge conflicts and failing CI builds, which aligns with this observation.

If this PR is still intended to land (perhaps it takes a different approach or covers additional cases), a few items to address:

- Merge conflicts with main need resolution
- CI build failures need investigation
- The TypeDeclaration.tsx change (adding `kind="type"`) seems unrelated to the peer/dev deps feature and could be split into its own commit or removed if #347 already covers it
- Second commit message "changes" is non-descriptive

PR #392, opened 2026-04-22
CSharp handle all names in name policy (escape invalid ones with @)

timotheeguerin:
Have the c# name policy handle invalid c# identifier (keywords or containing invalid char). Use `@` to escape
Expose helpers


bterlson:
Implemented as part of #396.

PR #393, opened 2026-04-23
Docs: Walk through: Fix model tsx file name

t20:
The model component has a code snippet, but the title of the code component is wrong and can be misleading.


bterlson:
Thank you!!

PR #394, opened 2026-04-23
Deduplicate names after name policy is applied

timotheeguerin:
depends on #394


packages/core/src/symbols/symbol-table.ts
```diff
@@ -13,7 +13,7 @@ export abstract class SymbolTable extends ReactiveUnionSet<OutputSymbol> {
   #deconflictNames = () => {
     for (const name of this.#namesToDeconflict) {
       const conflictedSymbols = [...this].filter(
-        (sym) => sym.originalName === name && !sym.ignoreNameConflict,
+        (sym) => sym.name === name && !sym.ignoreNameConflict,
```
timotheeguerin, line 16: @bterlson do you know why this used to only dedup with the original name?

bterlson:
Implemented as part of #396.

PR #396, opened 2026-04-23
fix(typescript): break long imports, clean up imports/symbols on re-render

bterlson:
Fixes two long-standing issues with TypeScript import handling:

## 1. Long imports now break across lines

Previously, `<ImportStatement>` concatenated named-bindings into a flat string, so `import { a, b, c, ... } from "..."` rendered as one long line regardless of `printWidth`. The statement is now wrapped in a `<group>` with `<line>`/`<indent>`/`<ifBreak>` so the layout engine can break it:

```ts
import {
  alphaItem,
  bravoItem,
  charlieItem,
  deltaItem,
  echoItem,
  foxtrotItem,
} from "testLib";
```

Short imports still render as `import { a, b } from "..."` on one line.

## 2. Imports and symbols no longer leak on re-render

The original bug: wrap an external refkey in a `<Show>`, toggle it off, and the `import` statement stays in the file. Toggle a locally-conflicting `<VarDeclaration>` on/off repeatedly and the local drifts `foo` → `foo_2` → `foo_3` because the stale import keeps participating in name-conflict resolution.

Root cause: `ref()`'s call to `addImport` had no inverse, and `createValueSymbol`/`createTypeSymbol` did not register `onCleanup`. The name-conflict resolver also never revisited aliases when a conflicting symbol was deleted.

Changes:

- **`TSModuleScope.addImport` is now reference-counted.** A new `removeImport` decrements the count; when it hits zero, the local import symbol is deleted and its `ImportedSymbol` / `ImportRecords` entries are pruned.
- **`ref()` registers an `onCleanup`** next to each `addImport` call (both the cross-package and cross-module paths) so the refcount balances 1:1 with live memo runs.
- **`createValueSymbol` / `createTypeSymbol` / `createTypeAndValueSymbol` now register `onCleanup`** to delete the symbol on component unmount. This closes a long-standing leak affecting every declaration component that uses them (`<VarDeclaration>`, function/class/interface declarations, etc.).
- **`SymbolTable` now re-runs conflict resolution on delete** by scheduling deconfliction for the deleted symbol's `originalName`. This lets survivors reclaim the plain name.
- **Both the default conflict handler and `tsNameConflictResolver` reset auto-generated aliases** (only those matching `${originalName}_N`) back to the plain name when the conflict resolves, leaving manually-set names untouched.

## Tests

- New `packages/typescript/test/imports-reactive.test.tsx` with 3 tests:
  - Removing the only reference via `<Show>` removes the import statement.
  - Repeated `<Show>` toggles don't drift name-conflict aliases (no `foo_3`/`foo_4`).
  - Multiple `<Show>` blocks referencing the same import — unmounting one keeps the import, unmounting all removes it.
- `packages/typescript/test/imports.test.tsx` + `externals.test.tsx` updated with formatting tests at `printWidth={40}` and adjustments for imports that legitimately exceed 80 columns.
- All 309 `@alloy-js/typescript` tests + 353 `@alloy-js/core` tests pass.

Chronus changeset: `.chronus/changes/ts-imports-fix-2026-4-23.md`.


bterlson:
Flagging interaction with #394 for reviewers:

The new `onDelete → re-schedule deconflict` hook in this PR pushes `symbol.originalName` into the deconflict queue, matching the existing filter (`sym.originalName === name`). Per the design intent discussed in https://github.com/alloy-framework/alloy/pull/394#discussion_r3131227073 (`originalName` → determinism, no "pump til stable"), this is the correct symmetry.

If #394 lands (filter swap to `.name === name`), my `onDelete` here would stop working for the winner-deletion case: pushing `originalName` (e.g. `"foo"`) would no longer match any survivor whose `.name` is `"foo_2"`, so the alias would never revert.

Resolving the tension is a #394 concern: under a `.name`-keyed filter, `onDelete` needs to re-gather the `originalName` cohort and push each survivor's current `.name` (or similar). Both this hook and #394's filter change are incremental hacks on top of the same inadequate foundation that the proposed global-conflict-handler rewrite would replace.

bterlson:
Implemented the canonicalName unification on top of the `deconflictedName` slot (commit bd0da761).

**Change:** `SymbolTable` now groups conflicts by `canonicalName(sym) = namePolicy(originalName)` (or `originalName` when no policy / `ignoreNamePolicy` is set). `onAdd`, `onDelete`, and the `#deconflictNames` filter all key on this same value.

**Why this subsumes #394:** the filter previously compared raw `originalName`, so `foo_bar` and `fooBar` under camelCase silently rendered to the same name without deconfliction. Grouping by canonical name closes that gap. It also fixes a pre-existing asymmetry where `onAdd` pushed `.name` (post-policy) while the filter matched on `originalName` (pre-policy).

**Why it doesn't pump:** canonical name is derived purely from immutable `originalName` + `namePolicy`. Resolver renames write `deconflictedName` (changing `.name` but not canonicalName), so the next pass finds the same cohort and terminates — matching @timotheeguerin's determinism requirement from #394's review thread.

Added two tests covering the camelCase `foo_bar`/`fooBar` dedup and the corresponding survivor-revert when the winner is deleted. All 1388 tests pass across core/ts/python/csharp/java/go.

cc @timotheeguerin — if this looks good I think #394 can close in favor of this.

PR #402, opened 2026-04-30
Replace render tree with AlloyNode printer

bterlson:
Previously Alloy rendered components into an array-based `RenderedTextTree`, converted that tree into Prettier's document IR, and then let Prettier produce final text. This PR replaces that path with a DOM-like `AlloyNode` tree: JSX children are inserted into nodes directly, `<SourceFile>`/`<SourceDirectory>` are represented as tree elements with context metadata, and a direct AlloyNode printer formats source output without building a separate print tree.

- Replace RenderedTextTree/Prettier-doc output with a DOM-like AlloyNode tree, direct printer, and tree-associated diagnostics/finalization.
- Public API: tree helpers now operate on AlloyNode; node classes/constants/factories are root exports; getContextForRenderNode remains as a deprecated alias; old RenderedTextTree/PrintHook APIs are removed; PrintTreeOptions no longer includes test-only noFlush.
- Update docs/tests for the final architecture, including runtime reconciliation and render diagnostics coverage.

Perf vs origin/main:
- render-long-file: 201ms vs 304ms (-34% CPU), mem -26%.
- render-many-files: 3.27s vs 4.12s (-21% CPU), mem -29%.
- ts-huge-object: 6.74s vs 9.25s (-27% CPU), mem -35%.
- render-imports: new printer-heavy benchmark, 7.83s.
- large codegen project: render/print 27.6s vs 50.9s (-46%).


packages/core/package.json
```diff
@@ -85,8 +85,11 @@
     "ws": "catalog:"
   },
   "devDependencies": {
+    "@alloy-js/babel-preset": "workspace:~",
     "@alloy-js/cli": "workspace:~",
     "@alloy-js/rollup-plugin": "workspace:~",
+    "@babel/core": "catalog:",
+    "@babel/preset-typescript": "catalog:",
```
timotheeguerin, line 91: Do we need all those now?

packages/core/src/render/get-string-width.ts
```diff
@@ -0,0 +1,201 @@
...
+    (x >= 129340 && x <= 129349) ||
+    (x >= 129351 && x <= 129535) ||
+    (x >= 129648 && x <= 129660) ||
+    (x >= 129664 && x <= 129673) ||
+    (x >= 129679 && x <= 129734) ||
+    (x >= 129742 && x <= 129756) ||
+    (x >= 129759 && x <= 129769) ||
+    (x >= 129776 && x <= 129784) ||
+    (x >= 131072 && x <= 196605) ||
+    (x >= 196608 && x <= 262141)
```
timotheeguerin, line 168: Fun code 😂

packages/core/src/render/node.ts
```diff
@@ -0,0 +1,440 @@
...
+export function createElement(localName: string, data?: unknown): ElementNode {
+  return new ElementNode(localName, data);
+}
+
+export function createFragment(): FragmentNode {
+  return new FragmentNode();
+}
+
+// ─────────────────────────────────────────────────────────────────────────────
+// Eager break-propagation
```
timotheeguerin: Nit should those use #region

packages/core/src/debug/render.test.tsx
```diff
@@ -106,90 +112,47 @@ it("sends render tree messages during render", async () => {
...
-  await collector.waitForFlush();
+  // Text content from Foo's body is also exposed as text nodes.
+  const values = nodeAdded.map((m) => m.value);
+  expect(values).toContain("Hello");
+  expect(values).toContain("World");
+});
 
-  expect(renderCount).toBe(2);
-  collector.stop();
+it.skip("rerenders when devtools requests rerender", async () => {
```
timotheeguerin: On purpose?

packages/core/src/render/node.ts
```diff
@@ -0,0 +1,440 @@
...
+ * Anything that can be inserted: a Node, a string (auto-wrapped as Text),
+ * or a number (auto-wrapped as Text).
+ */
+export type Insertable = AlloyNode | string | number;
+
+// ─────────────────────────────────────────────────────────────────────────────
+// Base Node — tree pointers + modern mutation API
+// ─────────────────────────────────────────────────────────────────────────────
+
+export abstract class AlloyNode {
```
timotheeguerin, line 36: Is class based polymorphism the best option here?

packages/core/package.json
bterlson: Nope I'll remove.

packages/core/src/render/get-string-width.ts
bterlson: Yessir!

packages/core/src/debug/render.test.tsx
bterlson: Yeah working to restore this functionality.

packages/core/src/render/node.ts
bterlson: Sure.

packages/core/src/render/node.ts
bterlson: Eh, going for DOM node parity, tagged objects are always a good option too ofc.

PR #403, opened 2026-05-06
Create PydanticClassDeclaration

mauriciogardini:
## Summary

Adds **Pydantic v2–oriented codegen** to **`@alloy-js/python`**: a dedicated **`PydanticClassDeclaration`** component, richer **`pydantic`** / **`typing`** / **`pydantic_settings`** externals, and **`decorators`** support on functions and methods so stacked decorators (e.g. Pydantic validators) emit in the right order above **`@classmethod`** / **`@staticmethod`**.

### Component & configuration
- **`PydanticClassDeclaration`** — Renders `class …(BaseModel):` with optional **`modelConfig`** (structured `ConfigDict` kwargs, camelCase props mapped to snake_case) or **`modelConfigExpression`** for arbitrary `model_config = …` when you need full control.
- Defaults to **`BaseModel`** when **`bases`** is omitted; supports explicit bases (e.g. **`RootModel`**).

### Builtins / externals (`packages/python/src/builtins/python.ts`)
- **`pydanticModule`** — Root symbols used with Pydantic models (e.g. **`BaseModel`**, **`Field`**, **`ConfigDict`**, validators/serializers, **`RootModel`**, **`SecretStr`**, etc.), plus **`alias_generators`** and **`types`** submodules.
- **`typingModule`** — Common **`typing`** symbols for annotations.
- **`pydanticSettingsModule`** — **`BaseSettings`** / **`SettingsConfigDict`** for settings-style models.

### Functions & methods
- **`CommonFunctionProps` / `BaseFunctionDeclaration`** — Optional **`decorators`** emitted before `def`.
- **`MethodDeclarationBase`** — Composes user **`decorators`** with intrinsic behavior (e.g. **`abstractmethod`**).
- **`ClassMethodDeclaration` / `StaticMethodDeclaration`** — Emit **`decorators`** first, then **`@classmethod`** / **`@staticmethod`**, then the method body (so e.g. **`@field_validator`** appears above **`@classmethod`**).

### Tests & changelog
- **`pydanticclassdeclarations.test.tsx`** — Broad coverage: models, imports, **`model_config`**, **`Field`**, inheritance, **`SecretStr`**, **`RootModel`**, forward refs / **`FutureStatement`**, alias generators, settings, **`computed_field`** / **`model_validator`**, **`typing`** / **`pydantic.types`**, etc. (aligned with current **`TYPE_CHECKING`** import emission on **`main`**).
- **`methoddeclaration.test.tsx`** — Adjustments for decorator ordering behavior.
- **Chronus** — `.chronus/changes/feature-python-pydantic-2026-5-5.md` for **`@alloy-js/python`**.


packages/python/src/builtins/python.ts
```diff
@@ -32,3 +32,56 @@ export const requestsModule = createModule({
     models: ["Response", "Request"],
   },
 });
+
+export const pydanticModule = createModule({
```
swatkatz, line 539: how did we come up with this list?

packages/python/src/builtins/python.ts
```diff
@@ -32,3 +32,56 @@ export const requestsModule = createModule({
...
+      "field_validator",
+      "model_serializer",
+      "model_validator",
+    ],
+    alias_generators: ["to_camel", "to_pascal", "to_snake"],
+    types: ["PositiveFloat", "PositiveInt"],
+  },
+});
+
+/** `typing` stdlib symbols used heavily with Pydantic generics and annotations. */
```
swatkatz, line 544: same question as above :)

packages/python/src/components/ClassMethodDeclaration.tsx
```diff
@@ -17,18 +18,31 @@ import { MethodDeclarationBase } from "./MethodBase.js";
  * def create(cls, value: str) -> None:
  *     return cls(value)
  * ```
+ *
+ * @remarks
+ * Use **`decorators`** for decorators that must appear above `@classmethod`
```
swatkatz, line 24: Is there an order in which decorators should be applied?

packages/python/src/builtins/python.ts
mauriciogardini: The list isn't an exhaustive one, it's what was judged by Cursor as being the minimum needed to emit idiomatic Pydantic v2 model classes, validators, serializers, and aliasing.

Do we want to be exhaustive, maybe?

packages/python/src/builtins/python.ts
mauriciogardini: Yeah, same as above:

The list isn't an exhaustive one, it's what was judged by Cursor as being the minimum needed to emit idiomatic Pydantic v2 model classes, validators, serializers, and aliasing.

Do we want to be exhaustive, maybe?

packages/python/src/components/ClassMethodDeclaration.tsx
mauriciogardini: I created a decorators component that enforces the correct order

packages/python/src/builtins/python.ts
timotheeguerin: doesn't need to be part of this PR but is there a way to dynamically create this (with a python script loading the modules for exmaple?)

packages/python/test/decoratorlist.test.tsx
```diff
@@ -0,0 +1,95 @@
...
+/**
+ * Direct contract tests for `DecoratorList`. The component is exercised
+ * transitively by every method/property test, but pinning the three core
+ * behaviors here makes regressions caught at the helper's own boundary —
+ * not after they propagate through every consumer.
+ */
+describe("DecoratorList", () => {
+  it("renders nothing when decorators is undefined", () => {
+    expect(
+      toSourceText([
```
timotheeguerin, line 15: can we use `expect().toRenderTo` instead for those?

packages/python/src/components/ClassMethodDeclaration.tsx
```diff
@@ -17,18 +18,24 @@ import { MethodDeclarationBase } from "./MethodBase.js";
...
+ * (for example Pydantic `@field_validator`).
  */
 export interface ClassMethodDeclarationProps extends CommonFunctionProps {
   abstract?: boolean;
 }
 
 export function ClassMethodDeclaration(props: ClassMethodDeclarationProps) {
   const sym = createMethodSymbol(props.name, { refkeys: props.refkey });
+  const { decorators, ...rest } = props;
```
timotheeguerin: I think you want to use alloy split helper here, spreading can cause issues with reactiveness

packages/python/src/components/PydanticClassDeclaration.tsx
```diff
@@ -0,0 +1,140 @@
...
+import type { ClassDeclarationProps } from "./ClassDeclaration.js";
+import { ClassDeclaration } from "./ClassDeclaration.js";
+
+/**
+ * Keyword-style options for Pydantic v2 `ConfigDict`, using camelCase prop names
+ * that map to snake_case Python arguments (for example `validateAssignment` →
+ * `validate_assignment`).
+ */
+export interface PydanticModelConfigDictProps {
+  arbitraryTypesAllowed?: boolean;
```
timotheeguerin, line 23: doc comments on the properties?

packages/python/src/components/PydanticClassDeclaration.tsx
```diff
@@ -0,0 +1,140 @@
...
+import { Atom } from "./Atom.jsx";
+import type { ClassDeclarationProps } from "./ClassDeclaration.js";
+import { ClassDeclaration } from "./ClassDeclaration.js";
+
+/**
+ * Keyword-style options for Pydantic v2 `ConfigDict`, using camelCase prop names
+ * that map to snake_case Python arguments (for example `validateAssignment` →
+ * `validate_assignment`).
+ */
+export interface PydanticModelConfigDictProps {
```
timotheeguerin, line 19: I know nothing of pydantic so just throwing this as a generic question but could it make sense to have those properites promoted at the top level so you can use like

```tsx
<PydanticClassDeclaration frozen>...
```

mauriciogardini:
Pydantic module generated (but filtering out v1) with:

```
#!/usr/bin/env node

import { execFileSync } from "node:child_process";
import { mkdirSync, writeFileSync } from "node:fs";
import { dirname, resolve } from "node:path";
import { fileURLToPath } from "node:url";

const __filename = fileURLToPath(import.meta.url);
const __dirname = dirname(__filename);

const args = process.argv.slice(2);
const pythonArgIndex = args.indexOf("--python");
const pythonExecutable =
  pythonArgIndex >= 0 && args[pythonArgIndex + 1] ?
    args[pythonArgIndex + 1]
  : "python3";

const pythonCode = String.raw`
import importlib
import json
import pkgutil
import re
import sys

try:
    import pydantic
except Exception as ex:
    print(f"Could not import pydantic: {ex}", file=sys.stderr)
    sys.exit(2)

identifier = re.compile(r"^[A-Za-z_][A-Za-z0-9_]*$")

def public_names(module):
    names = getattr(module, "__all__", None)
    if not names:
        names = [name for name in dir(module) if not name.startswith("_")]

    result = []
    for name in names:
        if isinstance(name, str) and identifier.match(name):
            result.append(name)
    return sorted(set(result))

descriptor = {".": public_names(pydantic)}

for module_info in pkgutil.iter_modules(pydantic.__path__):
    short_name = module_info.name
    if short_name.startswith("_"):
        continue
    full_name = f"pydantic.{short_name}"
    try:
        module = importlib.import_module(full_name)
    except Exception:
        continue
    names = public_names(module)
    if names:
        descriptor[short_name] = names

output = {
    "version": getattr(pydantic, "__version__", "unknown"),
    "descriptor": dict(sorted(descriptor.items())),
}
print(json.dumps(output, ensure_ascii=True))
`;

const raw = execFileSync(pythonExecutable, ["-c", pythonCode], {
  cwd: __dirname,
  encoding: "utf8",
  stdio: ["ignore", "pipe", "pipe"],
});

const parsed = JSON.parse(raw);
const descriptor = parsed.descriptor;
const pydanticVersion = parsed.version;

const outputFile = resolve(
  __dirname,
  "src/builtins/pydantic-descriptor.generated.ts",
);

mkdirSync(dirname(outputFile), { recursive: true });

const descriptorText = JSON.stringify(descriptor, null, 2);
const fileContents = `/* eslint-disable */
/**
 * This file is auto-generated by \`packages/python/generate-pydantic-builtins.mjs\`.
 * Source: installed pydantic package (version ${pydanticVersion}).
 * Do not edit by hand.
 */
import type { ModuleDescriptor } from "../create-module.js";

export const pydanticDescriptor = ${descriptorText} as const satisfies ModuleDescriptor;
`;

writeFileSync(outputFile, fileContents, "utf8");
console.log(`Wrote ${outputFile}`);
```

packages/python/src/builtins/python.ts
mauriciogardini: Used a Python script to create all relevant imports, except v1 ones.

timotheeguerin (approved):

PR #405, opened 2026-05-21
feat(typespec): add interface declarations and operations

glecaros:
Add InterfaceDeclaration and InterfaceOperationDeclaration components to the TypeSpec package with support for:
- Template parameters (simple and constrained)
- Interface extends clauses
- Operation parameters, return types, and 'is' syntax
- Name policy and deconfliction
- Refkey-based cross-declaration references

Extract shared Parameters component from operation-declaration to avoid duplication between OperationDeclaration and InterfaceOperationDeclaration.

Clean up test files: remove unused imports and fix nested afterEach anti-pattern.


packages/typespec/src/components/interface/interface-operation.tsx
```diff
@@ -0,0 +1,62 @@
...
+export interface InterfaceOperationDeclarationProps {
+  name: string | Namekey;
+  refkey?: Refkey;
+  templateParameters?: (string | TemplateParameterDescriptor)[];
+  parameters?: ParameterDescriptor[];
+  returnType?: Children;
+  is?: Children;
+}
+
+export function InterfaceOperationDeclaration(
```
timotheeguerin: do we need to have an interfaceOperation? could the Operation declaration detect if it is in an interface or not and include the right `op` prefix?

glecaros: I'll merge them... I had a reason when I first made this change but can't remember (which I take is a sign that it wasn't a good reason :) )

timotheeguerin (approved):

PR #408, opened 2026-05-25
feat(typespec): add decorator application and augment decorator compo…

glecaros:
…nents

- Add DecoratorApplication component for @decorator(args) syntax
- Add AugmentDecorator component for @@decorator(target, args) syntax
- Add decorators prop to all decoratable types: ModelDeclaration, ModelProperty, EnumDeclaration, EnumMember, UnionDeclaration, UnionVariant, OperationDeclaration, ScalarDeclaration
- Add tests for decorator rendering on each declaration type

### DIFF
--- a/packages/typespec/src/components/decorator/decorator-application.tsx
+++ b/packages/typespec/src/components/decorator/decorator-application.tsx
@@ -1,0 +1,45 @@
+import { Children, For, Indent } from "@alloy-js/core";
+
+export interface DecoratorApplicationProps {
+  /** The decorator name or a Reference to a decorator symbol. */
+  decorator: Children;
+
+  /** Arguments to pass to the decorator. */
+  args?: Children[];
+}
+
+/**
+ * Renders a TypeSpec decorator application.
+ *
+ * @example
+ * ```tsx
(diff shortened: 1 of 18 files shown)


packages/typespec/src/components/decorator/decorator-application.tsx
```diff
@@ -0,0 +1,45 @@
...
+            <Indent softline trailingBreak>
+              <For each={props.args} comma line>
+                {(arg) => arg}
+              </For>
+            </Indent>
+            )
+          </>
+        )}
+      </group>
+      <hbr />
```
timotheeguerin, line 42: do we want to force decorator to be applied on a new line each? I know it's a tricky thing to resolve in alloy.

glecaros: This does not seem unreasonable as seems to be behavior that `tsp format` has.

timotheeguerin: well technically the formatter preseve it on the same line if there is 1 or 2 and they don't overflow the line max width.

timotheeguerin (approved):

PR #409, opened 2026-05-27
feat(typespec): add AliasDeclaration component

glecaros:
Add AliasDeclaration component for rendering TypeSpec alias declarations:
  alias Name<T> = Type

Features:
- Template parameters with constraints and defaults
- Name policy support (quotes reserved words)
- Symbol deconfliction within namespaces
- Cross-namespace reference resolution with automatic using generation


packages/typespec/src/components/alias/alias-declaration.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+it("renders an alias with template parameters", () => {
+  expect(
+    <Output namePolicy={createTypeSpecNamePolicy()}>
+      <SourceFile path="main.tsp">
+        <AliasDeclaration name="Response" templateParameters={["T"]} type="T" />
+      </SourceFile>
+    </Output>,
+  ).toRenderTo({
+    "main.tsp": `
```
timotheeguerin, line 59: nit: you don't needw to specify a file if there is a single output, just pass the expected output of the only file. Makes the tests a little cleaner

timotheeguerin (approved):

packages/typespec/src/components/alias/alias-declaration.test.tsx
glecaros: I'll update in an upcoming PR to avoid the need for a refresh approve :)

PR #410, opened 2026-05-29
feat(typespec): add expression components and model property defaults

glecaros:
New components:
- ArrayExpression: renders Type[] array syntax
- IntersectionExpression: renders A & B intersection syntax
- SpreadExpression: renders ...Type spread syntax for models and enums

Enhanced:
- ModelProperty: add optional 'default' prop for default values


packages/typespec/src/components/array/array-expression.tsx
```diff
@@ -0,0 +1,22 @@
+import { Children } from "@alloy-js/core";
+
+export interface ArrayExpressionProps {
+  /** The element type of the array. */
+  type: Children;
```
timotheeguerin, line 5: should that be children?

glecaros: I think it should for stuff [like this](https://typespec.io/playground/?e=%40typespec%2Fopenapi3&c=bW9kZWwgRm9vIHsNCiAgYmFyOsYKxAx6OiBzdHJpbmc7xBJ9W107DQp9&options=%7B%7D&vs=%7B%7D)

packages/typespec/src/components/array/array-expression.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+        <ModelDeclaration name="Page">
+          <ModelProperty
+            name="items"
+            type={<ArrayExpression type="string" />}
+          />
+        </ModelDeclaration>
+      </SourceFile>
+    </Output>,
+  ).toRenderTo({
+    "main.tsp": `
```
timotheeguerin: shouldn't need the file, also fine if you want to merge this and do a final cleanup when we get this into main

timotheeguerin (approved):

packages/typespec/src/components/array/array-expression.test.tsx
glecaros: fixed :)

PR #412, opened 2026-05-29
feat(typespec): add const declarations and value expressions

glecaros:
Add ConstDeclaration, ObjectValue, ObjectValueProperty, and ArrayValue components for TypeSpec const declarations and value expressions.

- ConstDeclaration: renders const with optional type annotation
- ObjectValue/ObjectValueProperty: renders #{ key: value } syntax
- ArrayValue: renders #[item1, item2] syntax
- Added 'const' to NamedTypeKind and name policy

### DIFF
--- a/packages/typespec/src/components/const/const-declaration.test.tsx
+++ b/packages/typespec/src/components/const/const-declaration.test.tsx
@@ -17,11 +17,9 @@ it("renders a const declaration with a value", () => {
         <ConstDeclaration name="maxAge" value="255" />
       </SourceFile>
     </Output>,
-  ).toRenderTo({
-    "main.tsp": `
+  ).toRenderTo(`
       const maxAge = 255
-    `,
-  });
+  `);
 });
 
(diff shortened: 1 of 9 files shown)


packages/typespec/src/components/const/const-declaration.test.tsx
```diff
@@ -0,0 +1,89 @@
...
+    <Output namePolicy={createTypeSpecNamePolicy()}>
+      <SourceFile path="main.tsp">
+        <ConstDeclaration name="maxAge" value="255" />
+      </SourceFile>
+    </Output>,
+  ).toRenderTo({
+    "main.tsp": `
+      const maxAge = 255
+    `,
+  });
```
timotheeguerin:
```suggestion
  ).toRenderTo(`
      const maxAge = 255
  `);
```
should work for all the test as theere is a single file

timotheeguerin (approved):

PR #415, opened 2026-06-05
feat(typespec): add scalar initializers, fix name policy, export miss…

glecaros:
…ing components

- Add ScalarInitializer component with doc/decorators support
- Update ScalarDeclaration to accept children for body content
- Replace incomplete keyword list with full TypeSpec keywords from compiler
- Fix name policy to properly detect invalid identifiers (not just 5 words)
- Export Reference, Parameters, and ScalarDeclaration from components index
- Export components, contexts, and name-policy from package index


packages/typespec/src/components/scalar-declaration/scalar-initializer.tsx
```diff
@@ -0,0 +1,37 @@
...
+ * @example
+ * ```tsx
+ * <ScalarInitializer name="fromInt" parameters={[{ name: "value", type: "uint32" }]} />
+ * ```
+ * This will produce:
+ * ```typespec
+ * init fromInt(value: uint32);
+ * ```
+ */
+export function ScalarInitializer(props: ScalarInitializerProps) {
```
timotheeguerin: technically called `ScalarConstructor` in typespec, probably better to allign

glecaros: fixed, and created https://github.com/microsoft/typespec/pull/11032 too :)

timotheeguerin (approved):

PR #417, opened 2026-06-08
feat(typespec): add StringLiteral and StringTemplateExpr components

glecaros:
- StringLiteral renders single-line (`"..."`) or multi-line (`"""..."""'`) string literals via the multiline prop
- StringTemplateExpr wraps children in ${} for interpolation
- Multi-line strings use raw <indent> intrinsic to keep closing triple quotes at the same indentation level as content
- Export both components from package entry point

### DIFF
--- a/packages/typespec/src/components/string-literal/string-literal.test.tsx
+++ b/packages/typespec/src/components/string-literal/string-literal.test.tsx
@@ -11,13 +11,26 @@ beforeEach(() => {
   resetProgram();
 });
 
-it("renders a simple string literal", () => {
+it("renders a simple string literal via value prop", () => {
+  expect(<StringLiteral value="hello world" />).toRenderTo(`"hello world"`);
+});
+
+it("renders a simple string literal via children", () => {
   expect(<StringLiteral>hello world</StringLiteral>).toRenderTo(
     `"hello world"`,
   );
 });
 
-it("renders a multi-line string literal", () => {
+it("renders a multi-line string via value prop", () => {
+  expect(<StringLiteral value={"line one\nline two"} multiline />).toRenderTo(`
+    """
+      line one
+      line two
+      """
+  `);
+});
+
+it("renders a multi-line string via children", () => {
   expect(
     <StringLiteral multiline>
       {code`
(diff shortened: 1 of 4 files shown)


packages/typespec/src/components/string-literal/string-literal.tsx
```diff
@@ -0,0 +1,69 @@
+import { Children } from "@alloy-js/core";
+
+export interface StringLiteralProps {
+  /**
+   * The string content. May include {@link StringTemplateExpr} children
+   * for interpolation.
+   */
+  children: Children;
```
timotheeguerin: Wondering here if it would not work better with an attribute. I feel like with multiline we might also want to make sure you can do something like

```ts
<StringLiteral value="abc\ndef" multiline>
```
and it would render 
```tsp
"""
abc
def
"""
```

timotheeguerin: wondering if actually shouldn't be the default

packages/typespec/src/components/string-literal/string-literal.test.tsx
```diff
@@ -0,0 +1,78 @@
...
+it("renders a multi-line string literal", () => {
+  expect(
+    <StringLiteral multiline>
+      {code`
+        line one
+        line two
+      `}
+    </StringLiteral>,
+  ).toRenderTo(`
+    """
```
timotheeguerin, line 48: this feels a little weird as the default format but I assume this would look different in an expression, can we add a test with const or alias

packages/typespec/src/components/string-literal/string-literal.tsx
glecaros: I updated the behavior to be take both `value` and `children`. If using value, and the string contains a new line we use multiline (single otherwise). If using children, we default to single (as it probably makes sense if adding StringTemplateExpr.

packages/typespec/src/components/string-literal/string-literal.test.tsx
glecaros: with the updated value param seems better, added a test using in an alias.

timotheeguerin (approved):

PR #438, opened 2026-07-07
fix(typespec): isolate model expression scopes to prevent property de…

glecaros:
…confliction

ModelExpression was not creating its own scope, so ModelProperty symbols from sibling expressions shared the parent NamedTypeScope and were incorrectly deconflicted (e.g. statusCode_2, body_2).

Create an anonymous NamedTypeScope in ModelExpression so each expression isolates its own property symbols.

### DIFF
--- a/packages/typespec/src/components/model/model-expression.tsx
+++ b/packages/typespec/src/components/model/model-expression.tsx
@@ -1,9 +1,20 @@
-import { Block, Children } from "@alloy-js/core";
+import { Block, Children, onCleanup, Scope, useScope } from "@alloy-js/core";
+import { NamedTypeScope } from "../../scopes/named-type.js";
+import { NamedTypeSymbol } from "../../symbols/named-type.js";
 
 export interface ModelExpressionProps {
   children?: Children;
 }
 
 export function ModelExpression(props: ModelExpressionProps) {
-  return <Block>{props.children}</Block>;
+  const parentScope = useScope();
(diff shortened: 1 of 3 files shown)


packages/typespec/src/components/model/model-expression.tsx
```diff
@@ -1,9 +1,20 @@
...
 export interface ModelExpressionProps {
   children?: Children;
 }
 
 export function ModelExpression(props: ModelExpressionProps) {
-  return <Block>{props.children}</Block>;
+  const parentScope = useScope();
+  const sym = new NamedTypeSymbol("{anonymous}", undefined, "model");
+  onCleanup(() => sym.delete());
+  const scope = new NamedTypeScope(sym, parentScope);
```
timotheeguerin, line 12: this seems a little weird to have that name scope here, shouldn't we have something closer to interface expression in ts lib?

glecaros: At least in the current design, we would need to introduce another symbol type/scope to manage this (we would need a symbol that can hold members so essentially a copy of NamedTypeSymbol) and a scope type to contain it... if you think it's better, I can definitely do it. In the meantime, I moved the symbol creation to its own helper to make what we are doing a bit clearer and marked the symbol as transient.

timotheeguerin (approved):

timotheeguerin:
Something still feels off but if that works lets just merge that for now and we can revisit later

